PR #10, opened 2023-03-24
lint should fail when deprecated code ise used

StojanovicMilos:


package.json
```diff
@@ -41,6 +41,7 @@
     "@typescript-eslint/eslint-plugin": "5.51.0",
     "@typescript-eslint/parser": "5.51.0",
     "eslint": "8.33.0",
```
typhoon41, line 43: Please revert this change. This is the wrong package.json. You need to add it to projects/eslint-config one.
On top of that, you need to add this to both dependencies and devDependencies.

StojanovicMilos: Done

typhoon41 (approved):

PR #11, opened 2023-03-28
All libraries are now up to date with all the rules specified.

typhoon41:
Deprecated rules will be removed under different ticket once we move to Prettier.


TeaEntry (approved):
looks good to me :)

PR #12, opened 2023-03-29
Fixed broken stylelint.

typhoon41:


projects/stylelint-config/index.js
```diff
@@ -6,8 +6,8 @@ module.exports = {
   plugins: [
     "stylelint-scss",
     "stylelint-use-nesting",
-    "stylelint-selector-no-empty",
-    "stylelint-group-selectors",
+    //"stylelint-selector-no-empty",
```
TeaEntry, line 9: can you add a comment what was breaking or remove these lines?

typhoon41: Comment is in package.json. Didn't want to duplicate justification, but I'll add something.

TeaEntry (approved):

PR #13, opened 2023-03-30
Features/eslint upgrade

typhoon41:


projects/eslint-config/package.json
```diff
@@ -13,46 +13,44 @@
...
     "eslint-plugin-prefer-arrow": "1.2.3",
     "eslint-plugin-promise": "6.1.1",
     "eslint-plugin-unused-imports": "2.0.0",
     "eslint-plugin-xss": "0.1.12",
-    "@angular-eslint/builder": "14.0.2",
-    "@angular-eslint/eslint-plugin": "14.0.2",
-    "@angular-eslint/eslint-plugin-template": "14.0.2",
-    "@angular-eslint/schematics": "14.0.2",
-    "@angular-eslint/template-parser": "14.0.2"
+    "@angular-eslint/builder": "14.4.0",
```
msljivic, line 28: Perhaps we should consider if we need to specify @angular-eslint/xxx dependencies or not. Points to consider

- Do we support single angular version?
- Is the major version of @angular-eslint is aligned with the major angular version (@angular-eslint packages v14 and @angular v14 are compatible)?
- How is this 'compatibility' implied to users of the library. (Angular v13 users can install the library and have potential issues - there is no peer dependency on angular major version.)

I've had a look at the similar project, https://github.com/Tinkoff/linters. They don't include the @angular-eslint dependencies.

msljivic (approved):

PR #51, opened 2023-05-22
BP-527:  Move entry search filter to entry-components library.

msljivic:


markovic-katarina (approved):
TODO: Add Config class with defaults for e.g. maxLength and add applyButtonText etc.

PR #59, opened 2023-05-26
BP-540: Initial implementation of Docs application containing dialog & header components

mrle:


angular.json
```diff
@@ -122,6 +122,123 @@
           }
         }
       }
+    },
+    "@enigmatry/docs": {
```
msljivic, line 126: Maybe we should rename project name to something like @enigmatry/docs-app (just project name, nothing else)
so it's clear when building with cli: ng build @enigmatry/docs-app

msljivic (approved):

PR #81, opened 2023-06-01
Replace node_modules relative path when importing sass-true.

msljivic:
We should avoid using a relative path to node_modules in use declarations, because the location of node_modules can differ depending on install location.

(sass-true is used by `layout/_grid.scss` and `text/_modification.scss` making it a peer-dependency of the scss-foundation library)


typhoon41:
Yeah, this is ideal and code was working like this in past, but I remember we had some issues after sass-true upgrade and we had to make it peer dependency.

Have you executed all the commands (such as builds, linting, testing) before creating this PR?

msljivic:
> Have you executed all the commands (such as builds, linting, testing) before creating this PR?

Yes.

typhoon41 (approved):

PR #85, opened 2023-06-05
Update Angular to v14.

msljivic:


mrle (approved):
@msljivic what abut codegen genrated forms? Should they be update, if not already, to be in Angular 14+ format?

msljivic:
Not sure, there a few options, currently we generate 
```form = new FormGroup({})```
this is a typed form group (where type is inferred go empty object :))

we could use UntypedFormGroup or FormGroup<T>

PR #88, opened 2023-06-13
BP-547: Introduce static/dynamic select filter in search filter component

mrle:


libs/entry-components/search-filter/public-api.ts
```diff
@@ -1,8 +1,11 @@
...
 
-export { SearchFilterParams } from './search-filter-params';
-export { SearchFilterBase } from './search-filter-base';
-export { SearchFilterInput } from './search-filter-textbox';
+export { SearchFilterParams } from './search-filter-params.type';
+export { SearchFilterBase } from './search-filter-input/search-filter-base.model';
+export { TextSearchFilter } from './search-filter-input/inputs/text-search-filter.model';
+export { SelectSearchFilter } from './search-filter-input/inputs/select-search-filter.model';
+export { DynamicSelectSearchFilter } from './search-filter-input/inputs/dynamic-select-search-filter.model';
+export { SelectSearchFilterOption } from './search-filter-input/inputs/select-search-filter-option.model';
```
mrle, line 8: Regarding this model. We have SelectOption inside entry-form component that can be re-used here as well, but it is in different lib.

Would it make sense to re-use autocomplete component here?

libs/entry-components/search-filter/search-filter-input/search-filter-input.component.html
```diff
@@ -0,0 +1,27 @@
...
+
+    <mat-select *ngSwitchCase="controlType.select"
+      [formControlName]="searchFilter.key" [id]="searchFilter.key" [multiple]="searchFilter.multiSelect">
+      <mat-option *ngIf="!searchFilter.multiSelect" [value]="undefined">
+        {{config.noneSelectedOptionText}}
+      </mat-option>
+      <mat-option *ngFor="let option of searchFilter.options" [value]="option.key">{{option.label}}</mat-option>
+    </mat-select>
+
+    <mat-select *ngSwitchCase="controlType.dynamicSelect"
```
mrle, line 18: we could use entry-autocomplete komponent here but then we need to add reference to entry-form?

mrle: If we add this ref then we could re-use SelectOption which will answer my other question

mrle: we agreed not to re-use autocomplete from entry-form lib

msljivic (approved):

PR #96, opened 2023-06-16
BP-569: Merging header styles from Blueprint with component

Cvetanovic:
https://jira.enigmatry.com/browse/BP-569


apps/demo-app/src/styles/modules/theming/_theming.scss
```diff
@@ -3,22 +3,22 @@
...
-			hover-background: vars.$accent,
-			hover-font-color: vars.$font-color-light,
-		),
-		button-accent: (
-			default-background: vars.$accent,
-			default-font-color: vars.$font-color-light,
-			hover-background: vars.$primary,
-			hover-font-color: vars.$font-color-light,
-		)
+		// // header: (
```
mrle, line 6: fell free to remove theis commented in header theme config. It will not be in use anymore, right?

PR #98, opened 2023-06-20
BP-569:Merging header styles from Blueprint with component

Cvetanovic:
https://jira.enigmatry.com/browse/BP-569


typhoon41 (changes requested):

apps/demo-app/src/app/examples/header/header/header-example.component.html
```diff
@@ -1,15 +1,15 @@
 <entry-header [title]="'Users'">
-    <div buttons class="action-buttons header-item">
-        <button class="button-accent item" (click)="addNew()">Add new</button>
-        <button class="button-primary item" (click)="otherAction()">Other action</button>
+    <div buttons class="action-buttons item">
+        <button class="entry-button-accent button first-button" (click)="addNew()">Add new</button>
+        <button class="entry-button-primary button" (click)="otherAction()">Other action</button>
     </div>
```
typhoon41, line 5: To conclude, after removing clickable class, buttons still have cursor effect?

libs/entry-components/header/styles/partials/core/components/headers/_general.scss
```diff
@@ -1,25 +1,14 @@
 @use '../../../../../../styles/modules/variables' as vars;
-// @use '../../../../../../styles/modules/breakpoints' as breakpoints;
 
 .entry-header {
-	padding: 4px vars.$large-gap vars.$default-gap;
+	padding: 24px vars.$default-gap;
+	background: vars.$primary-lightest;
```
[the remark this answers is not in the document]
typhoon41: This is unresolved. In partials we can only have hard-coded styles, without variables. Everything that gets value from theme, must be in adequate generate mixin.

libs/entry-components/header/styles/modules/components/headers/_general.scss
```diff
@@ -7,6 +7,7 @@
 
         .title {
             font-size: map.value-from($theme, 'header.title-size');
+            color: map.value-from($theme, 'header.title-color');
```
typhoon41, line 10: You changed the structure of object. This won't work anymore => 'header.title-color'
You also need to change it to =>'header.title.color' in order to work. Same for size.

PR #104, opened 2023-06-27
BP-585: Tweak headers to support theming and use shared styles

Cvetanovic:
https://jira.enigmatry.com/browse/BP-585


typhoon41 (changes requested):

libs/scss-entry/src/modules/_default-theme.scss
```diff
@@ -1,5 +1,12 @@
 /* stylelint-disable number-max-precision */
 $theme: (
+	header: (
+        background-color: #DFF1FD,
+		title: (
+            size: 32px,
+            color: #2581C4
```
typhoon41, line 7: In fact, this should be primary color

PR #108, opened 2023-07-05
Features/bp 570 forms theming

typhoon41:


libs/entry-form/styles/partials/core/components/forms/_general.scss
```diff
@@ -1,13 +1,26 @@
...
-            margin-left: 10px;
+		.button {
+			margin-left: 10px;
 
-            &:first-child {
-                margin-left: 0;
-            }
-        }
-    }
+			&:first-child {
```
Cvetanovic, line 8: is it possible to change &:first-child with some class name (first-button, or first-item) as we have in the header, or that here is impossible?

typhoon41: Yes, we can probably do that but we'll have ground refactoring ticket once we switch to angular v16 + code was untouched this time.

Cvetanovic (approved):

PR #130, opened 2023-08-03
BP-621: Fix issues related to forms

Cvetanovic:
https://jira.enigmatry.com/browse/BP-621


libs/entry-components/styles/modules/_default-theme.scss
```diff
@@ -41,8 +42,9 @@ $theme: (
 			checkmark-color: #000,
 			selected-border-color: #000,
 			selected-background-color: #FFF,
-			border: 2px solid #0000008A,
-			border-hover: 2px solid #0000008A,
+			border-width: 2px,
+			border-color: #0000008A,
```
jovana-marceta, line 46: @Cvetanovic  Can that color be replaced with shorthand hexadecimal value?

PR #138, opened 2023-08-16
BP-619: Fix issues related to v15 migration regarding look & feel on table component

Cvetanovic:


libs/entry-components/styles/partials/core/components/tables/_action-cells.scss
```diff
@@ -1,16 +1,13 @@
...
 .entry-table-checkbox-cell {
-	flex: none;
-	width: 56px;
-	max-width: 40px;
+	width: 46px;
 
 	.mat-mdc-checkbox {
-		display: flex;
+		.mdc-form-field {
+			margin-bottom: 0;
```
typhoon41, line 6: Why do we need this margin? I don't think we need this.

libs/entry-components/styles/partials/core/components/tables/_layout.scss
```diff
@@ -1,22 +1,8 @@
...
-
-		&.mat-mdc-table-with-data {
-			flex: 1;
-		}
-	}
-
-	.mat-mdc-table {
-		min-width: 100%;
-		border-collapse: separate;
+.mat-table-container {
```
typhoon41, line 3: Please leave table wrapper (or in case of other components, other entry- prefix on the first class) since this might have more effect than we'd like.

libs/entry-components/styles/partials/core/components/tables/_action-cells.scss
Cvetanovic: because in forms/general we have .mdc-form-field with margin-bottom: 10px. Checkbox also has the same class name inside it, it is not aligned with the text in the columns (it goes up 10px)

typhoon41: That's for forms. It isn't applied on table since table isn't in form. As you can see...

image

PR #141, opened 2023-08-23
Features/bp 650 table simplification

typhoon41:


libs/entry-table/src/lib/components/entry-table/entry-table.component.html
```diff
@@ -1,11 +1,7 @@
-<div class="entry-table-main entry-table-layout">
-
+<div #tableContainer class="entry-table">
```
msljivic, line 1: class="entry-table" is removed from the host tag and added to the first wrapping div.

I see that classes are added by the code generator as well, unfortunately. Code generator adds two classes "entry-table" and "entry-list-name-table" e.g.
<entry-table class="entry-table entry-products-table "
</entry-table>

We could remove the host tag class from the code generator as well. This could be a breaking change for styling in existing projects.
Or we can use a different class for the wrapping div, and keep the host tag classes.
What do you think?

typhoon41: I'd then keep host tag class and that code gen behavior, but then I'd remove wrapping div and instead of defining #tableContainer, I'd just get host tag element and scroll it in TS logic.
Does that sound OK?

msljivic: Yes, good idea.

msljivic (approved):

PR #144, opened 2023-08-28
Fix issues on Dialog and Search Filter

typhoon41:
https://jira.enigmatry.com/browse/BP-606


apps/demo-app/src/app/examples/button/button-example.module.ts
```diff
@@ -2,26 +2,17 @@ import { NgModule } from '@angular/core';
...
   ],
   imports: [
     CommonModule,
-    SharedModule,
-    EntryButtonModule
+    SharedModule
   ],
   exports: [
     ButtonExampleComponent
-  ],
```
msljivic, line 18: The example is incomplete without this code.

typhoon41: I agree and disagree in the same time. We need to have definition for buttons in shared module since that's optimal and representative; developers will look at this app as example for many best practices, that's why we can't allow inline styles and other crap.
On the other side I partially agree since in theory we can use these buttons independently, but there's not too much sense in that if we only support submit and cancel button. Whenever we have those buttons we usually (should) have a form and ideally generated one which already includes buttons.

msljivic (approved):

avanwaas (approved):

PR #153, opened 2023-10-23
Theme findings fix

typhoon41:
Inputs can now have configurable background color. Cell padding on tables are now configurable through theming.

### DIFF
--- a/libs/entry-components/styles/_generator.scss
+++ b/libs/entry-components/styles/_generator.scss
@@ -2,6 +2,7 @@
 @use 'modules/components/forms/generator' as forms;
 @use 'modules/components/search-filter/generator' as search-filter;
 @use 'modules/components/tables/generator' as table;
+@use 'modules/components/inputs/generator' as inputs;
 @use 'modules/components/buttons/generator' as buttons;
 @use 'modules/default-theme' as default;
 @use 'modules/vendors/angular-material/generator' as material-theme;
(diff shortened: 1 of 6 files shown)


libs/entry-components/styles/_generator.scss
```diff
@@ -2,6 +2,7 @@
 @use 'modules/components/forms/generator' as forms;
 @use 'modules/components/search-filter/generator' as search-filter;
 @use 'modules/components/tables/generator' as table;
```
jovana-marceta, line 4: I assume that 'dialog' and 'table' modules are not in the plural form because they accurately represent singular entities within theming?

typhoon41: They should be plural. Will be fixed.

libs/entry-components/styles/modules/components/dialogs/_generator.scss
```diff
@@ -2,12 +2,12 @@
 
 @mixin generate-from($theme) {
 	.entry-dialog {
-		.title {
+		.dialog-header .title {
```
Cvetanovic, line 5: is it possible to put the title inside dialog-header, with that will be 3 nesting levels, or I'm wrong?

typhoon41: It's 3 levels one way or another :)

Cvetanovic (approved):

jovana-marceta (approved):

PR #154, opened 2023-10-25
BP-508: Disable submit button on submit to prevent users from submitting more than once

msljivic:


msljivic:
How to create custom event modifiers
 * https://github.com/Tinkoff/ng-event-plugins,
 * https://github.com/angular/angular/blob/main/packages/platform-browser/src/dom/events/key_events.ts

vsljivic (approved):

PR #172, opened 2023-11-30
Fixes after brainstorming rules with TLs.

typhoon41:


typhoon41:
- padded-blocks => Padded blocks seems fine since it prevents redundant empty space in beginning and end of each block. It is also auto-fixable. Any side effects you encountered with this rule? Indeed deprecated, on master is already replaced with stylistic corresponding rule.
- func-style => Why would you use function at all nowadays? With variable as arrow function you can make it readonly and code is more encapsulated. I think people should actually learn to code that way the same way we're not using public setters anymore in C#. Why would you use function before defining it in typescript? Aren't we using typescript to normalize this quirky behaviour of JS?
- @typescript-eslint/no-unused-vars => we can decrease severity of this to warning so you can develop without blockages, but you should still fix this before pushing final solution. It's rightfully one of eslint recommended rules turned on by default.
- @typescript-eslint/no-magic-numbers => I feel pain regarding this I'm not against it, but we can perhaps keep it by loosening it by allowing hard-coded ArrayIndexes, default values of objects, initial class field values (options were added later)?
- @typescript-eslint/no-non-null-assertion => I think form.get('name') falls into 5% category we'd like to ignore. This rule helped us a lot in past to spot holes in our nswag-generated API related code + is forcing us (not) to have guards. Any other problematic and realistic example that is causing you pain you'd like to share with us?
- @typescript-eslint/space-before-function-paren => I think this shouldn't be changed since it's how community expects function to look like + it's autofixed.
- unused-imports/no-unused-imports => similar as for no-unused-vars (decrease severity to warning). Take into account we'll put treshold of X warnings after which build's gonna be red.
- no-negated-condition => legit comment, but is it the case in more than 5% of cases?
- no-warning-comments => this has warning severity, so it doesn't break builds nor blocks anyone. Totally fine to add TODOs to code base, but if you have hundreds of them in your code base (seen cases), something's gotta give. We can talk about treshold value, but this reminds us to address them and treshold reminds us that if they pile up we gotta do something about them for real.
- no-loops => get this part, but should we left warning for people to actually learn something and reconsider using linq?

avanwaas (approved):

PR #174, opened 2023-12-26
BP-549: As a user I want a spinner while data is loading on a page.

msljivic:


libs/entry-components/loader/entry-loader/loader.component.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.createOverlay();
+    this.overlayRef.attach(new TemplatePortal(this.templateRef, this.viewContainerRef));
+  }
+
+  ngOnDestroy(): void {
+    this.disposeOverlayRef();
+  }
+
+  private createOverlay() {
+    const overlayConfig = new OverlayConfig({
```
zsrdjan, line 51: What is all this needed? Because mat-spinner does not create an overlay, or because of some other reason.

apps/demo-app/src/app/examples/loader/loader-example/loader-example.component.html
```diff
@@ -0,0 +1,31 @@
+<section>
+  <entry-loader *ngIf="loading" [fullscreen]="fullscreen" [color]="color" [backgroundOverlay]="backgroundOverlay"></entry-loader>
```
zsrdjan, line 2: Maybe name it enableBackgroundOverlay

libs/entry-components/loader/entry-loading-bar/loading-bar.component.ts
```diff
@@ -0,0 +1,33 @@
...
+
+@Component({
+  selector: 'entry-loading-bar',
+  templateUrl: './loading-bar.component.html',
+  styleUrls: ['./loading-bar.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class EntryLoadingBarComponent {
+
+  @Input() type: 'loader' | 'progress-bar' = 'loader';
```
zsrdjan, line 15: spinner | progres-bar

zsrdjan (approved):

PR #175, opened 2024-01-05
BP-788: Theming: add letter spacing property

Cvetanovic:
https://enigmatry.atlassian.net/browse/BP-788

### DIFF
--- a/libs/entry-components/styles/modules/_default-theme.scss
+++ b/libs/entry-components/styles/modules/_default-theme.scss
@@ -16,19 +16,23 @@ $theme: (
 		fonts: (
 			hero-titles: (
 				family: '',
-				size: 0
+				size: 0,
+				letter-spacing: null
 			),
 			titles: (
 				family: '',
-				size: 0
+				size: 0,
+				letter-spacing: null
 			),
 			body: (
 				family: '',
-				size: 0
+				size: 0,
+				letter-spacing: null
 			),
 			buttons: (
 				family: '',
-				size: 0
+				size: 0,
+				letter-spacing: null
 			)
 		),
 		spacing: (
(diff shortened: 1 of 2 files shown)


libs/entry-components/styles/modules/vendors/angular-material/_typography-generator.scss
```diff
@@ -23,6 +23,12 @@
 	$typography: -set-font-size-for('titles', $typography, $fonts, ('subtitle-1', 'subtitle-2'));
 	$typography: -set-font-family($typography, $fonts);
```
typhoon41, line 25: -set-font-size-for and -set-letter-spacing-for are identical methods, so they should be joined in the following way:
1) get $letter-spacing from theme like we do for fonts
2) create third method called -set-font-for with the same parameters as for set-font-size-for and inject $letter-spacing to fourth argument
3) In that new method c/p code from -set-font-size-for and on the first line get letter-spacing as you did in -set-letter-spacing-for
4) First do @each and inside @each add ifs and optionally add font-size and letter-spacing if desired values aren't null

libs/entry-components/styles/modules/_default-theme.scss
```diff
@@ -16,19 +16,23 @@ $theme: (
...
+				letter-spacing: null
 			),
 			titles: (
 				family: '',
-				size: 0
+				size: 0,
+				letter-spacing: null
 			),
 			body: (
 				family: '',
```
typhoon41, line 28: Have you tested if these changes work? You can execute tests with new properties.

typhoon41 (changes requested):

libs/entry-components/styles/modules/_default-theme.scss
Cvetanovic: letter-spacing and font-size have been tested and both of them work properly after the last changes

jovana-marceta (approved):

PR #179, opened 2024-01-16
BP-788: Update documentation with new letter space property

jovana-marceta:
Extending documentation with new letter spacing property that is included. Check out the details in this [pull request](https://github.com/enigmatry/entry-angular-building-blocks/commit/b66c29b11bd4b4cc3db9b38404fa8f9b6cab7388).

### DIFF
--- a/libs/entry-components/configure-theming.md
+++ b/libs/entry-components/configure-theming.md
@@ -272,11 +272,11 @@ Allow customisation of typography and font styles based on Angular Material typo
 
 Each font related property in configuration can have: **family**, **size** and **letter-spacing** values.
 
-- `body`: Define typography for base body text.
-  - `family`: Set custom font family
-  - `size`: Specify a custom font size.
-  - `letter-spacing`:  Customize the letter spacing;  set to `null` by default, inheriting Angular Material's values.
-- `buttons`: Typography for buttons and anchors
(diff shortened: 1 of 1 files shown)


libs/entry-components/configure-theming.md
```diff
@@ -269,9 +269,13 @@ $custom-theme: (
 ### 3. Fonts
 
 Allow customisation of typography and font styles based on Angular Material typography levels. `body` property is the most important and **mandatory** to set because that one is **default** and font family will be applied on all headers and buttons. If we want to additionally customize some other typography elements and properties, we can include other font properties.
-Each font related property in configuration can have **family** and **size** values.
+
+Each font related property in configuration can have: **family**, **size** and **letter-spacing** values.
 
 - `body`: Define typography for base body text.
+  - `family`: Set custom font family
```
typhoon41, line 276: We're using set, specify and customize words to effectively describe the same action in one group of properties.
We also want to be consistent in regard whether we use dot in the end of sentence or not.

i.e. `body`: Defines typography for base body text.
  - `family`: Sets custom font family
  - `size`: Sets custom font size
  - `letter-spacing`:  Sets custom letter spacing;  if we leave default value or set it to `null` manually, Angular Material's default values will be applied

PR #180, opened 2024-01-19
BP-787: Extend the entry-search-filter with datetime inputs

laslorajda:
BP-787


msljivic:
- Create Custom DateAdapter that injects existing DateAdapter and configures it for date-time display
- Provide this custom adapter in component provider scope
- Specify format and compare function via injection token (ENTRY_MAT_DATE_TIME)

libs/entry-components/search-filter/control-type.ts
```diff
@@ -1,5 +1,6 @@
 export enum ControlType {
     text = 'text-input',
     select = 'select-input',
-    autocomplete = 'autocomplete-input'
+    autocomplete = 'autocomplete-input',
+    dateTime = 'date-time-input'
```
msljivic, line 5: add date component as well

msljivic (approved):

PR #193, opened 2024-03-04
BP-950: Fix background color for inputs

Cvetanovic:


jovana-marceta (approved):

libs/entry-components/styles/modules/_default-theme.scss
```diff
@@ -45,6 +45,9 @@ $theme: (
 		inputs: (
 			background: null
 		),
+		paginators: (
```
jovana-marceta, line 48: Since paginators are inheriting background from inputs by default, maybe it should be implemented as submap for inputs instead of separate entity?

libs/entry-components/styles/modules/components/inputs/_generator.scss
```diff
@@ -11,5 +12,17 @@
 				border-radius: 4px;
 			}
 		}
+
```
typhoon41, line 15: What exactly you're trying to solve with this? Background color on focus? We probably want to skip that, since grayish color is everywhere on the focus and it's not the way.

libs/entry-components/styles/modules/_default-theme.scss
typhoon41: I'm also bit unsure whether this paginator thing is mature enough to end in theme. It's more logical for all fields to have the same color and we have only one exception for this logic in our projects so far.

libs/entry-components/styles/modules/_default-theme.scss
```diff
@@ -45,6 +45,9 @@ $theme: (
...
+			background: null
+		),
```
typhoon41, line 50: Documentation is not updated to go in line with these changes.

PR #194, opened 2024-03-06
BP-952: Set font size for table header cells

jovana-marceta:
Improving the theme structure with an additional submap dedicated to table headers, nabling the customization of font size and weight in header cells. 
https://enigmatry.atlassian.net/browse/BP-952

### DIFF
--- a/libs/entry-components/styles/modules/_default-theme.scss
+++ b/libs/entry-components/styles/modules/_default-theme.scss
@@ -51,8 +51,7 @@ $theme: (
 	),
 	tables: (
 		header: (
-			font-size: null,
-			font-weight: null
+			font-size: null
 		),
 		cells: (
 			edge-gap: 4px,
--- a/libs/entry-components/styles/modules/_default-theme.scss
+++ b/libs/entry-components/styles/modules/_default-theme.scss
@@ -52,7 +52,11 @@ $theme: (
 	tables: (
 		cells: (
 			edge-gap: 4px,
-			padding: null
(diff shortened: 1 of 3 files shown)


libs/entry-components/styles/modules/_default-theme.scss
```diff
@@ -50,6 +50,10 @@ $theme: (
 		)
 	),
 	tables: (
+		header: (
+			font-size: null,
+			font-weight: null
```
typhoon41, line 55: Do we really need to support weight in the end?

jovana-marceta: Perhaps it would be best to delay font-weight property for the time being, and revisit the idea if there is a need for it in future projects. Implementing it now would mean adding a property for the sorted state too, so it might be set aside for now if  @Cvetanovic  agrees?

libs/entry-components/styles/modules/components/tables/_header.scss
```diff
@@ -0,0 +1,9 @@
+@use 'sass:map';
+
+@mixin generate-from($theme) {
+	$header-cell-font-size: map.get($theme, 'tables', 'header', 'font-size');
```
Cvetanovic, line 4: Maybe you can put cell-font-size instead of header-cell-font-size, file is called header

typhoon41 (approved):

Cvetanovic (approved):

PR #195, opened 2024-03-07
BP-955: Set background color for radio buttons

jovana-marceta:
https://enigmatry.atlassian.net/browse/BP-955

### DIFF
--- a/libs/entry-components/.stylelintrc.json
+++ b/libs/entry-components/.stylelintrc.json
@@ -1,6 +1,3 @@
 {
-  "extends": "@enigmatry/stylelint-config",
-  "ignoreFiles": [
-    "styles/modules/components/buttons/_radio.scss"
-  ]
+  "extends": "@enigmatry/stylelint-config"
 }
--- a/libs/entry-components/.stylelintrc.json
+++ b/libs/entry-components/.stylelintrc.json
@@ -1,3 +1,6 @@
 {
-  "extends": "@enigmatry/stylelint-config"
+  "extends": "@enigmatry/stylelint-config",
+  "ignoreFiles": [
+    "styles/modules/components/buttons/_radio.scss"
+  ]
 }
--- a/libs/entry-components/styles/modules/components/buttons/_radio.scss
+++ b/libs/entry-components/styles/modules/components/buttons/_radio.scss
@@ -5,12 +5,10 @@ $radio-circles-selector: '.mdc-radio__outer-circle, .mdc-radio__inner-circle';
 
 @mixin generate-from($theme) {
 	.mat-mdc-radio-group .mat-mdc-radio-button {
-		$radio-border: map.get($theme, 'general', 'buttons', 'radio', 'background');
-
 		&.mat-mdc-radio-checked .mdc-radio {
 			&:not(:disabled):active #{$radio-selector}, #{$radio-selector} {
 				#{$radio-circles-selector} {
-					border-color: $radio-border;
+					border-color: map.get($theme, 'general', 'buttons', 'radio', 'background');
 				}
 			}
 		}
--- a/libs/entry-components/styles/modules/components/buttons/_radio.scss
+++ b/libs/entry-components/styles/modules/components/buttons/_radio.scss
@@ -1,12 +1,22 @@
 @use 'sass:map';
 
 @mixin generate-from($theme) {
-	.mat-mdc-radio-button {
+	.mat-mdc-radio-group .mat-mdc-radio-button {
 		.mdc-radio {
 			.mdc-radio__native-control:enabled:checked + .mdc-radio__background {
 				.mdc-radio__outer-circle, .mdc-radio__inner-circle {
-					border-color: map.get($theme, 'general', 'buttons', 'radio', 'background') !important;
+					border-color: map.get($theme, 'general', 'buttons', 'radio', 'background');
 				}
+			}
+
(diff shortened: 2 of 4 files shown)


libs/entry-components/styles/modules/components/buttons/_radio.scss
```diff
@@ -0,0 +1,13 @@
+@use 'sass:map';
+
+@mixin generate-from($theme) {
+	.mat-mdc-radio-button {
+		.mdc-radio {
+			.mdc-radio__native-control:enabled:checked + .mdc-radio__background {
+				.mdc-radio__outer-circle, .mdc-radio__inner-circle {
+					border-color: map.get($theme, 'general', 'buttons', 'radio', 'background') !important;
```
jovana-marceta, line 8: @typhoon41  Acknowledging that the use important is discouraged, I am unsatisfied with this approach and will seek a more suitable solution. One possibility is to increase specificity by introducing a custom class in the Formly generated files for radio buttons.

libs/entry-components/styles/modules/components/buttons/_radio.scss
```diff
@@ -1,13 +1,23 @@
...
+	.mat-mdc-radio-group .mat-mdc-radio-button {
 		.mdc-radio {
 			.mdc-radio__native-control:enabled:checked + .mdc-radio__background {
 				.mdc-radio__outer-circle, .mdc-radio__inner-circle {
-					border-color: map.get($theme, 'general', 'buttons', 'radio', 'background') !important;
+					border-color: map.get($theme, 'general', 'buttons', 'radio', 'background');
 				}
 			}
+
+			&:not(:disabled):active .mdc-radio__native-control:enabled:checked + .mdc-radio__background .mdc-radio__outer-circle {
```
Cvetanovic, line 12: You have duplicated selectors. You can create variables for them and then you don't need ignore files in stylelint

libs/entry-components/styles/modules/components/buttons/_radio.scss
```diff
@@ -1,13 +1,23 @@
...
+				border-color: map.get($theme, 'general', 'buttons', 'radio', 'background');
+			}
+		}
+
+		&.mat-mdc-radio-checked .mdc-radio .mdc-radio__native-control:enabled:checked + .mdc-radio__background {
+			.mdc-radio__outer-circle, .mdc-radio__inner-circle {
```
Cvetanovic, line 18: duplicated selector

libs/entry-components/styles/modules/components/buttons/_radio.scss
```diff
@@ -1,13 +1,23 @@
 @use 'sass:map';
 
 @mixin generate-from($theme) {
-	.mat-mdc-radio-button {
+	.mat-mdc-radio-group .mat-mdc-radio-button {
 		.mdc-radio {
 			.mdc-radio__native-control:enabled:checked + .mdc-radio__background {
 				.mdc-radio__outer-circle, .mdc-radio__inner-circle {
-					border-color: map.get($theme, 'general', 'buttons', 'radio', 'background') !important;
+					border-color: map.get($theme, 'general', 'buttons', 'radio', 'background');
```
Cvetanovic, line 8: create variable for this map

libs/entry-components/styles/modules/components/buttons/_radio.scss
```diff
@@ -0,0 +1,23 @@
...
+	.mat-mdc-radio-group .mat-mdc-radio-button {
+		.mdc-radio {
+			.mdc-radio__native-control:enabled:checked + .mdc-radio__background {
+				.mdc-radio__outer-circle, .mdc-radio__inner-circle {
+					border-color: map.get($theme, 'general', 'buttons', 'radio', 'background');
+				}
+			}
+
+			&:not(:disabled):active .mdc-radio__native-control:enabled:checked + .mdc-radio__background .mdc-radio__outer-circle {
+				border-color: map.get($theme, 'general', 'buttons', 'radio', 'background');
```
typhoon41, line 13: The same variable should be used on all places (eliminate duplication)

libs/entry-components/.stylelintrc.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "@enigmatry/stylelint-config"
-}
\ No newline at end of file
+  "extends": "@enigmatry/stylelint-config",
+  "ignoreFiles": [
+    "styles/modules/components/buttons/_radio.scss"
```
typhoon41, line 4: Why this?!

typhoon41 (changes requested):

libs/entry-components/.stylelintrc.json
jovana-marceta: I initially it to the stylelint ignore list because it generated five warnings, which I believed were too many to address directly in the file. However, upon further review, I realized that by using variables for selectors, I could effectively eliminate these warnings.

typhoon41: I understand, but we have hundreds of rules being applied on every single file. If we're gonna turn off remaining 100 rules just for sake of 5, half of the solution would end up in ignore section.

Ignore section should be used exclusively (temporarily) for legacy code we don't have budget to deal with, vendor libraries (we're not responsible for their code quality) and other types of classes we can't have full control over (i.e. code generation or some PoC work).

typhoon41 (approved):

PR #197, opened 2024-03-14
BP-953: Set background for selected row in table

jovana-marceta:
Setting of a background color for selected row in table (if transparent is given, respect odd/even settings - ignore). Default setting should still be primary color to avoid breaking changes.

https://enigmatry.atlassian.net/browse/BP-953

### DIFF
--- a/libs/entry-components/styles/modules/_default-theme.scss
+++ b/libs/entry-components/styles/modules/_default-theme.scss
@@ -50,9 +50,6 @@ $theme: (
 		)
 	),
 	tables: (
-		header: (
-			font-size: null
-		),
 		cells: (
 			edge-gap: 4px,
 			padding: null
@@ -61,7 +58,8 @@ $theme: (
 			selected-color: #FFF,
 			disabled-color: #F5F5F5,
 			odd-even-row: odd,
-			odd-even-background: #F0F0F0
+			odd-even-background: #F0F0F0,
+			odd-even-background-selected: transparent
 		),
 		contents: (
 			no-result: (
--- a/libs/entry-components/styles/modules/_default-theme.scss
+++ b/libs/entry-components/styles/modules/_default-theme.scss
@@ -62,7 +62,7 @@ $theme: (
 			disabled-color: #F5F5F5,
 			odd-even-row: odd,
 			odd-even-background: #F0F0F0,
-			odd-even-background-selected: transparent
+			odd-even-background-selected: null
 		),
 		contents: (
 			no-result: (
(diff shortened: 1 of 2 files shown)


libs/entry-components/styles/modules/_default-theme.scss
```diff
@@ -61,7 +61,8 @@ $theme: (
 			selected-color: #FFF,
 			disabled-color: #F5F5F5,
 			odd-even-row: odd,
-			odd-even-background: #F0F0F0
+			odd-even-background: #F0F0F0,
+			odd-even-background-selected: null
```
typhoon41, line 65: This is actually selected-background? Odd-even has less to do with it.

libs/entry-components/styles/modules/components/tables/_rows.scss
```diff
@@ -2,6 +2,8 @@
 @use '@enigmatry/scss-foundation/src/modules/lists/row-coloring' as list;
 
 @mixin generate-from($theme) {
+	$selected-row-background: map.get($theme, 'tables', 'rows', 'odd-even-background-selected');
```
typhoon41, line 5: Keep variables close to their usage, in other words, please move this between line 15 and 16.

libs/entry-components/styles/modules/_default-theme.scss
jovana-marceta: I renamed it to selected-background and to avoid confusion with selected color, I decided to nest them together.

libs/entry-components/styles/modules/_default-theme.scss
```diff
@@ -58,10 +58,13 @@ $theme: (
 			padding: null
 		),
 		rows: (
-			selected-color: #FFF,
 			disabled-color: #F5F5F5,
 			odd-even-row: odd,
-			odd-even-background: #F0F0F0
+			odd-even-background: #F0F0F0,
+			selected: (
```
typhoon41, line 64: This change will break all the projects overriding default selected foreground color. It's fine as long as you'll fix all those projects once this is merged because we don't wanna get into situation where some dev updates minor or build version of entry-components just to find out SCSS is not compiling.

jovana-marceta: Acknowledged, I will reverse changes to ensure stability and prevent any potential problems with SCSS compilation, selecting a safer strategy at this moment.

Cvetanovic (approved):

typhoon41 (approved):

PR #200, opened 2024-03-18
BP-956: Set background for slide toggles

jovana-marceta:
https://enigmatry.atlassian.net/browse/BP-956


libs/entry-components/styles/modules/components/slide-togglers/_generator.scss
```diff
@@ -0,0 +1,33 @@
...
+		$enabled-states: 'enabled', 'enabled:focus:not(:active)', 'enabled:active', 'enabled:hover:not(:focus):not(:active)';
+
+		$track-on-background: map.get($theme, 'general', 'slide-togglers', 'on', 'track', 'background');
+		$track-off-background: map.get($theme, 'general', 'slide-togglers', 'off', 'track', 'background');
+		$handle-on-background: map.get($theme, 'general', 'slide-togglers', 'on', 'handle', 'background');
+		$handle-off-background: map.get($theme, 'general', 'slide-togglers', 'off', 'handle', 'background');
+
+		@each $state in $enabled-states {
+			.mdc-switch:#{$state} {
+				& .mdc-switch__track::after {
```
Cvetanovic, line 14: unnecessary spacing

libs/entry-components/styles/modules/components/slide-togglers/_generator.scss
```diff
@@ -0,0 +1,33 @@
...
+					background: $track-on-background;
+				}
+
+				& .mdc-switch__track::before {
```
Cvetanovic, line 18: unnecessary spacing

libs/entry-components/styles/modules/components/slide-togglers/_generator.scss
```diff
@@ -0,0 +1,33 @@
+/* stylelint-disable scss/selector-no-redundant-nesting-selector */
```
Cvetanovic, line 1: Please check one more time, I think you don't need this

Cvetanovic (changes requested):

libs/entry-components/styles/modules/components/slide-togglers/_generator.scss
jovana-marceta: It is done by purpose, It can't be separated, since mdc-switch__track is child element.

libs/entry-components/styles/modules/_default-theme.scss
```diff
@@ -54,7 +54,25 @@ $theme: (
 		checkboxes: (
 			background: null,
 			border-hover-color: null
-		)
+		),
+		slide-togglers: (
```
typhoon41, line 58: I'd name it just toggle. Simpler, shorter and removes unnecessary specificity.

libs/entry-components/styles/modules/components/slide-togglers/_generator.scss
```diff
@@ -0,0 +1,33 @@
...
+@use 'sass:map';
+@mixin generate-from($theme) {
+	.mat-mdc-slide-toggle {
+		$enabled-states: 'enabled', 'enabled:focus:not(:active)', 'enabled:active', 'enabled:hover:not(:focus):not(:active)';
+
+		$track-on-background: map.get($theme, 'general', 'slide-togglers', 'on', 'track', 'background');
```
typhoon41, line 7: To reduce amount of duplication and strings count, you can say:
$toggle: map.get($theme, 'general', 'toggle');
$track-on-background: map.get($toggle, 'on', 'track');
etc.

libs/entry-components/styles/modules/components/slide-togglers/_generator.scss
```diff
@@ -0,0 +1,33 @@
...
+		$track-off-background: map.get($theme, 'general', 'slide-togglers', 'off', 'track', 'background');
+		$handle-on-background: map.get($theme, 'general', 'slide-togglers', 'on', 'handle', 'background');
+		$handle-off-background: map.get($theme, 'general', 'slide-togglers', 'off', 'handle', 'background');
+
+		@each $state in $enabled-states {
```
typhoon41, line 12: We want to make things configurable, everything fine.
I'm thinking whether we can configure single color or just two colors instead of four? What are the use cases?

In AM we have two colors and two lighter/darker variations of colors for track depending on lightness. Is it the same on our projects?
Can we define colors just for handle on/off and then calculate color for track which is 10% darker or lighter depending on lightness of given color?

typhoon41 (changes requested):

libs/entry-components/styles/modules/components/slide-togglers/_generator.scss
jovana-marceta: Good point, I agree that is better to get rid of too specific customization for now. Now we will have just on/off for handle. I created function that conditionally increases or decreases the lightness of a color based on the color's current lightness, unless the color is undefined or transparent.

libs/entry-components/styles/modules/components/toggle/_generator.scss
```diff
@@ -0,0 +1,45 @@
...
+}
+
+@mixin generate-from($theme) {
+	.mat-mdc-slide-toggle {
+		$enabled-states: 'enabled', 'enabled:focus:not(:active)', 'enabled:active', 'enabled:hover:not(:focus):not(:active)';
+		$handle-selector: '.mdc-switch__handle::after';
+		$track-selector: '.mdc-switch__track';
+
+		$toggle: map.get($theme, 'general', 'toggle');
+		$handle-on: map.get($toggle, 'on');
```
typhoon41, line 20: IMHO, we're not getting more readability (in this case) with $handle-on and $handle-off variables. I suggest merge them to $handle-on/off-background.

libs/entry-components/styles/modules/components/toggle/_generator.scss
```diff
@@ -0,0 +1,45 @@
...
+			.mdc-switch:#{$state} {
+				&.mdc-switch--selected #{$handle-selector} {
+					background-color: $handle-on-background;
+				}
+
+				&:not(.mdc-switch--selected) #{$handle-selector} {
+					background-color: $handle-off-background;
+				}
+
+				#{$track-selector}::after {
```
typhoon41, line 35: Instead of having track-selector variable, why not join after and before with 
```
.mdc-switch__track {
   &::after {
   }
   
   &::before {
   }
}
```

typhoon41 (changes requested):

typhoon41 (approved):

PR #202, opened 2024-03-19
BP-976: Update documentation for theming with new properties

jovana-marceta:
https://enigmatry.atlassian.net/browse/BP-976

### DIFF
--- a/libs/entry-components/styles/modules/_default-theme.scss
+++ b/libs/entry-components/styles/modules/_default-theme.scss
@@ -66,25 +66,27 @@ $theme: (
 		),
 		cells: (
 			edge-gap: 4px,
-			padding: null
+			padding: null,
 		),
 		rows: (
+			odd-even-background: (
+				odd: #F0F0F0, 
+				even: null,
+			),
 			selected-color: #FFF,
-			selected-background: null,
+			selected-background: #2581C4,
 			disabled-color: #F5F5F5,
-			odd-even-row: odd,
-			odd-even-background: #F0F0F0
 		),
 		contents: (
 			no-result: (
-				font-size: 13px,
-				font-weight: 500
+				size: 13px,
+				font-weight: 500,
+			),
+		),
+		dialogs: (
+			title: (
+				size: 20px
 			)
-		)
-	),
-	dialogs: (
-		title: (
-			size: 20px
 		)
 	)
 );
--- a/libs/entry-components/styles/modules/_default-theme.scss
+++ b/libs/entry-components/styles/modules/_default-theme.scss
@@ -71,10 +71,10 @@ $theme: (
 		rows: (
 			odd-even-background: (
 				odd: #F0F0F0, 
-				even: null,
+				even: #FFF,
 			),
 			selected-color: #FFF,
-			selected-background: #2581C4,
+			selected-background: null,
 			disabled-color: #F5F5F5,
 		),
 		contents: (
--- a/libs/entry-components/styles/modules/components/buttons/_radio.scss
+++ b/libs/entry-components/styles/modules/components/buttons/_radio.scss
@@ -4,12 +4,13 @@ $radio-selector: '.mdc-radio__native-control:enabled:checked + .mdc-radio__backg
 $radio-circles-selector: '.mdc-radio__outer-circle, .mdc-radio__inner-circle';
 
 @mixin generate-from($theme) {
-	.mat-mdc-radio-group .mat-mdc-radio-button {
-		&.mat-mdc-radio-checked .mdc-radio {
-			&:not(:disabled):active #{$radio-selector}, #{$radio-selector} {
-				$radio-border: map.get($theme, 'general', 'buttons', 'radio', 'background');
+	$radio-border: map.get($theme, 'general', 'buttons', 'radio', 'background');
 
-				@if $radio-border {
+	@if $radio-border {
+		.mat-mdc-radio-group .mat-mdc-radio-button {
+			&.mat-mdc-radio-checked .mdc-radio {
+				
+				&:not(:disabled):active #{$radio-selector}, #{$radio-selector} {
 					#{$radio-circles-selector} {
 						border-color: $radio-border;
 					}
--- a/libs/entry-components/styles/modules/components/buttons/_radio.scss
+++ b/libs/entry-components/styles/modules/components/buttons/_radio.scss
@@ -9,7 +9,7 @@ $radio-circles-selector: '.mdc-radio__outer-circle, .mdc-radio__inner-circle';
 			&:not(:disabled):active #{$radio-selector}, #{$radio-selector} {
 				$radio-border: map.get($theme, 'general', 'buttons', 'radio', 'background');
 
-				@if $radio-border and $radio-border != transparent {
+				@if $radio-border {
 					#{$radio-circles-selector} {
 						border-color: $radio-border;
 					}
--- a/libs/entry-components/styles/modules/components/buttons/_radio.scss
+++ b/libs/entry-components/styles/modules/components/buttons/_radio.scss
@@ -7,8 +7,12 @@ $radio-circles-selector: '.mdc-radio__outer-circle, .mdc-radio__inner-circle';
 	.mat-mdc-radio-group .mat-mdc-radio-button {
 		&.mat-mdc-radio-checked .mdc-radio {
 			&:not(:disabled):active #{$radio-selector}, #{$radio-selector} {
-				#{$radio-circles-selector} {
-					border-color: map.get($theme, 'general', 'buttons', 'radio', 'background');
+				$radio-border: map.get($theme, 'general', 'buttons', 'radio', 'background');
+
+				@if $radio-border and $radio-border != transparent {
+					#{$radio-circles-selector} {
+						border-color: $radio-border;
+					}
 				}
 			}
 		}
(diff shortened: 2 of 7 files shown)


libs/entry-components/theming-configuration.md
```diff
@@ -350,32 +352,93 @@ $custom-theme: (
...
+
+- `on`: Contains property for the toggle in its active state.
+- `off`: Contains property for the toggle in its inactive state.
+  - `background`: Sets the toggle's handler and track background color based on activated(on) or deactivated(off) state of toggle. If transparent is passed, it will be ignored and primary background will be applied.
+
+```scss
+$custom-theme: (
+  general: (
+    toggle: (
+      on: (
```
typhoon41, line 404: This is fine and it works fine, but seems our implementation is wrong when we don't specify this.
As far as I can see, default value is null, but we're unconditionally setting it in entry-components\styles\modules\components\toggle\_generator.scss

libs/entry-components/theming-configuration.md
```diff
@@ -350,32 +352,93 @@ $custom-theme: (
...
+        background: #333
+      )
     )
   )
 )
 ```
 
+### 7. Checkboxes
+
+Submap provides customization of checkboxes background and border color for **unchecked state**.
```
typhoon41, line 376: In similar fashion as for other sections, you can just start with "Provides customization..."

libs/entry-components/theming-configuration.md
```diff
@@ -350,32 +352,93 @@ $custom-theme: (
...
+
+- `background`:  sets the background color for checkbox.
```
typhoon41, line 378: This sets color for all checkboxes, so please use plural.

libs/entry-components/theming-configuration.md
```diff
@@ -350,32 +352,93 @@ $custom-theme: (
...
+- `border-hover-color`: sets the checkbox border color on hovering while its unchecked.
```
typhoon41, line 379: sets the border color of unchecked checkbox while in hover state

libs/entry-components/theming-configuration.md
```diff
@@ -350,32 +352,93 @@ $custom-theme: (
...
+
+```scss
+$custom-theme: (
+  general: (
+    checkboxes: (
+      background: blue,
```
typhoon41, line 385: This is fine and it works fine, but seems our implementation is wrong when we don't specify this.
As far as I can see, default value is null, but we're unconditionally setting it in entry-components\styles\modules\components\checkboxes\_generator.scss

libs/entry-components/theming-configuration.md
```diff
@@ -350,32 +352,93 @@ $custom-theme: (
...
+- `radio`:  Contains property for radio buttons
+  - `background`: sets the background color for the inner and outer circles of radio buttons
 
 ```scss
 $custom-theme: (
   general: (
     buttons: (
-      icon-size: 48px
+      icon-size: 48px,
+      radio: (
```
typhoon41, line 366: This is fine and it works fine, but seems our implementation is wrong when we don't specify this.
As far as I can see, default value is null, but we're unconditionally setting it in entry-components\styles\modules\components\buttons\_radio.scss

libs/entry-components/theming-configuration.md
```diff
@@ -350,32 +352,93 @@ $custom-theme: (
...
+  general: (
+    checkboxes: (
+      background: blue,
+      border-hover-color: darkblue
+    )
+  )
+)
+```
+
+### 7. Toggles
```
typhoon41, line 392: Toggles are number 8.

typhoon41 (changes requested):

libs/entry-components/styles/modules/components/tables/_header.scss
```diff
@@ -3,7 +3,10 @@
 @mixin generate-from($theme) {
 	$cell-font-size: map.get($theme, 'tables', 'header', 'font-size');
 
-	.mat-mdc-header-cell {
-		font-size: $cell-font-size;
+	/* stylelint-disable-next-line scss/at-if-no-null */
+	@if $cell-font-size != null {
```
typhoon41, line 7: Shorter syntax (that we already used) is just @if $cell-font-size {

libs/entry-components/styles/modules/components/buttons/_radio.scss
```diff
@@ -7,8 +7,12 @@ $radio-circles-selector: '.mdc-radio__outer-circle, .mdc-radio__inner-circle';
 	.mat-mdc-radio-group .mat-mdc-radio-button {
 		&.mat-mdc-radio-checked .mdc-radio {
 			&:not(:disabled):active #{$radio-selector}, #{$radio-selector} {
-				#{$radio-circles-selector} {
-					border-color: map.get($theme, 'general', 'buttons', 'radio', 'background');
+				$radio-border: map.get($theme, 'general', 'buttons', 'radio', 'background');
+
+				@if $radio-border and $radio-border != transparent {
```
typhoon41, line 12: Why's transparent important in this case?

jovana-marceta: I experimented with transparent values and found that the radio button becomes invisible. In other scenarios, we allow the default values to prevail, so it would be beneficial to address that aspect as well.

typhoon41: mkay, but we haven't implemented odd/even coloring if transparent is passed (as described in main ticket) in entry-components\styles\modules\components\tables\_rows.scss

The idea is NOT to use primary color if transparent is given but to use underlying color (odd or even, depending on row itself)

libs/entry-components/styles/modules/components/buttons/_radio.scss
```diff
@@ -7,8 +7,12 @@ $radio-circles-selector: '.mdc-radio__outer-circle, .mdc-radio__inner-circle';
 	.mat-mdc-radio-group .mat-mdc-radio-button {
 		&.mat-mdc-radio-checked .mdc-radio {
 			&:not(:disabled):active #{$radio-selector}, #{$radio-selector} {
-				#{$radio-circles-selector} {
-					border-color: map.get($theme, 'general', 'buttons', 'radio', 'background');
+				$radio-border: map.get($theme, 'general', 'buttons', 'radio', 'background');
+
```
typhoon41, line 11: For all these files with if check, if should go to the most upper layer possible.
Right now we're still generating CSS and selectors but without any properties. We shouldn't generate anything if value is null for all the checks.

libs/entry-components/styles/modules/components/toggle/_generator.scss
```diff
@@ -22,24 +22,30 @@
...
 				}
 
-				.mdc-switch__track {
-					&::after {
-						background-color: -adjust-lightness($handle-on-background);
-					}
-			
-					&::before {
-						background-color: -adjust-lightness($handle-off-background);
+				@if $handle-on-background and $handle-off-background {
```
typhoon41, line 37: Do we require both colors to be defined or is it enough just one for everything to work?
IMHO it's fine either way, but this logic is conflicting. First two ifs are allowing independent definition, but the last one requires both...

jovana-marceta: We require both of them to this be able to work. Now we have only one condition and I moved it up.

libs/entry-components/styles/modules/_default-theme.scss
```diff
@@ -66,25 +66,27 @@ $theme: (
 		),
 		cells: (
 			edge-gap: 4px,
-			padding: null
+			padding: null,
 		),
 		rows: (
+			odd-even-background: (
+				odd: #F0F0F0, 
+				even: null,
```
typhoon41, line 74: This shouldn't be null but whitish color from AM (again, not to have breaking changes).

libs/entry-components/styles/modules/components/tables/_rows.scss
```diff
@@ -1,25 +1,48 @@
+/* stylelint-disable max-nesting-depth */
 @use 'sass:map';
-@use '@enigmatry/scss-foundation/src/modules/lists/row-coloring' as list;
+
+@mixin row-coloring($color-map) {
```
typhoon41, line 4: You don't need this mixin. List utility from Scss Foundation have it already.
We don't need to check whether we have key or not, since default theme will always have some default values.

libs/entry-components/styles/modules/components/tables/_rows.scss
```diff
@@ -1,25 +1,48 @@
...
+			@include row-coloring(map.get($theme, 'tables', 'rows', 'odd-even-background'));
+		
 			&.selected {
-				background-color: map.get($theme, 'general', 'colors', 'primary');
 				$selected-row-background: map.get($theme, 'tables', 'rows', 'selected-background');
-				
-				@if $selected-row-background and $selected-row-background != transparent {
+		  
+				@if $selected-row-background == transparent {
+					&:nth-child(odd) {
```
typhoon41, line 29: Use mixin here from Scss Foundation.

libs/entry-components/styles/modules/_default-theme.scss
```diff
@@ -66,25 +66,27 @@ $theme: (
 		),
 		cells: (
 			edge-gap: 4px,
-			padding: null
+			padding: null,
 		),
 		rows: (
+			odd-even-background: (
+				odd: #F0F0F0,
```
typhoon41, line 73: Documentation needs to be altered too for this new moment.

typhoon41 (approved):

PR #204, opened 2024-03-27
BP-1053: Date time picker does not select current date time when no value is set

laslorajda:
BP-1053


msljivic:
On opening event we set following:
```ts
this.control.setValue(this.control.value || now);
datePicker._selected = this.control.value;
datePicker.startAt = datePicker._selected;
```
So, we would get the todays date, even if date picker is opened and closed without selecting date

What if we only select the date without updating the control value:

```ts
datePicker._selected = this.control.value ?? new
datePicker.startAt = datePicker._selected;
```

msljivic (approved):

PR #211, opened 2024-04-29
BP-1042: Update to Angular 17

laslorajda:


package.json
```diff
@@ -46,18 +46,18 @@
...
+    "@angular-eslint/eslint-plugin": "17.3.0",
+    "@angular-eslint/eslint-plugin-template": "17.3.0",
+    "@angular-eslint/schematics": "17.3.0",
+    "@angular-eslint/template-parser": "17.3.0",
+    "@angular/cli": "^17.3.6",
+    "@angular/compiler-cli": "^17.3.6",
+    "@angular/material-date-fns-adapter": "17.3.6",
     "@ckeditor/ckeditor5-angular": "^5.2.0",
     "@ckeditor/ckeditor5-build-classic": "^37.0.0",
     "@enigmatry/scss-foundation": "15.0.0-preview.5",
```
avanwaas, line 63: we're at 16.1 already, shouldn't we upgrade this 1?

.nx/cache/18.3.4-nx.win32-x64-msvc.node
avanwaas, line 1: what's this?

package.json
```diff
@@ -23,16 +23,16 @@
   },
   "private": true,
   "dependencies": {
```
avanwaas, line 25: some of (dev)dependencies should be upgraded, create a new app using the angular cli and compare

avanwaas (dismissed):
please create a new app using angular cli and compare. Check e.g. tsconfig.json (es2022 for everything, ...)

package.json
laslorajda: I think I got them all, I ignore the ckeditor ones(since they are going away) and eslint/formly ones(since I don't know enough about them).

package.json
laslorajda: This feels like it shouldn't even be here, it means we are depending on ourselves?

msljivic:
We need to update angular version in:

libs/eslint-config/package.json

libs/entry-form/package.json
libs/entry-form/readme.md

libs/entry-components/package.json
libs/entry-components/readme.md

[README.md]

package.json
msljivic: Not sure, I think we use these to test if theme and components styles compile (they reference scss-foundation)
I would leave it for now.

msljivic:
Check if we have any vulnerabilities (after npm i or npm audit)

msljivic (approved):

PR #219, opened 2024-05-13
BP-1042: Remove Dependency on Stylelint-Config

TamaraMarr:


package.json
```diff
@@ -37,14 +37,8 @@
     "highlight.js": "^11.9.0",
     "lodash-es": "^4.17.21",
     "markdown-it": "^14.1.0",
-    "postcss-scss": "4.0.6",
+    "postcss-scss": "4.0.9",
```
msljivic, line 40: is in both dependencies and dev Dependencies

package-lock.json
```diff
@@ -5758,150 +5410,6 @@
...
-      "dev": true,
-      "optional": true,
-      "os": [
-        "linux"
-      ],
-      "engines": {
-        "node": ">= 10"
-      }
-    },
-    "node_modules/@nx/nx-linux-x64-gnu": {
```
msljivic, line 5857: not sure why are these @nx / platform packages removed again - the build pipeline is running on linux and then fails because of this ...

msljivic: we regenerated package.lock last time to resolve this issue

msljivic (approved):

PR #231, opened 2024-07-26
BP-1194: As a user, I want the UI to be WCAG 2.1 compliant to ensure accessibility for all users

Cvetanovic:
https://enigmatry.atlassian.net/browse/BP-1194

### DIFF
--- a/apps/demo-app/src/app/app.component.html
+++ b/apps/demo-app/src/app/app.component.html
@@ -1,7 +1,7 @@
 <div class="docs-main">
   <mat-toolbar class="docs-header">
     <div class="doc-navigation">
-      <img class="logo header-element" src="../assets/images/entry_logo_small_transparent_material.png" alt="Entry logo">
+      <img class="logo header-element" src="assets/images/entry_logo_small_transparent_material.png" alt="Entry logo">
       <button mat-flat-button color="primary" class="header-element" [routerLink]="''">
         Components
       </button>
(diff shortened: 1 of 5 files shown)


apps/demo-app/src/app/app.component.html
```diff
@@ -1,7 +1,7 @@
 <div class="docs-main">
   <mat-toolbar class="docs-header">
     <div class="doc-navigation">
-      <img class="logo header-element" src="../assets/images/entry_logo_small_transparent_material.png">
+      <img class="logo header-element" src="../assets/images/entry_logo_small_transparent_material.png" alt="Entry logo">
```
jovana-marceta, line 4: You don't need to use relative paths for assets because they are already defined in the angular.json file

apps/demo-app/src/app/app.component.html
```diff
@@ -10,7 +10,7 @@
 
     <a mat-flat-button color="primary" target="_blank" class="header-element"
       href="https://github.com/enigmatry/entry-angular-building-blocks">
-      <img class="github-icon" src="../assets/images/github-circle-white-transparent.svg" />
+      <img class="github-icon" src="../assets/images/github-circle-white-transparent.svg" alt="GitHub white icon" />
```
jovana-marceta, line 13: The purpose of this icon is to link to the GitHub source repository for Entry packages. Maybe to consider changing the alt text provides more context about the purpose of the image, which is helpful for users who rely on screen readers such as "View Enigmatry source code repository."

apps/demo-app/src/app/app.component.html
Cvetanovic: good catch mate ;)

jovana-marceta (approved):

msljivic (approved):

PR #232, opened 2024-08-07
SCSS Unit Testing

Cvetanovic:


libs/scss-foundation/tests/borders/border-radius.tests.scss
```diff
@@ -0,0 +1,40 @@
...
+					/* stylelint-disable scss/declaration-nested-properties */
+					top-left-radius: 0;
+					top-right-radius: 0;
+					bottom-right-radius: 0;
+					bottom-left-radius: 0;
+					/* stylelint-enable scss/declaration-nested-properties */
+				}
+			}
+		}
+	}
```
typhoon41, line 21: We could add another test (just for practice sake, this is more than enough for coverage): to test only provision of second and third argument.

Basically, to avoid calling function with many optional parameters, we can try syntax without nulls and with named params such as:
@include border.partial-border-radius($top-right: 10px, $bottom-right: 1.2em);

libs/scss-foundation/tests/borders/border-radius.tests.scss
```diff
@@ -20,6 +20,25 @@
...
+		  @include true.expect() {
+			border: {
+			  /* stylelint-disable scss/declaration-nested-properties */
+			  top-left-radius: 0;
+			  top-right-radius: 10px;
+			  bottom-right-radius: 1.2em;
+			  bottom-left-radius: 0;
+			  /* stylelint-enable scss/declaration-nested-properties */
+			}
+		  }
```
typhoon41, line 37: Formatting is wrong or what? Stylelint not working?

typhoon41 (approved):

PR #233, opened 2024-08-12
BP-1197: SCSS-Foundation: Add mixins for font size and weight

jovana-marceta:
https://enigmatry.atlassian.net/browse/BP-1197

We've noticed a frequent need to set font size and weight consistently across various projects, particularly in those lacking entry theming or with unique structures. To simplify our CSS and reduce redundancy, we suggest creating a single, adaptable mixin to manage both font size and weight.

### DIFF
--- a/libs/scss-foundation/src/modules/typography/_fonts.scss
+++ b/libs/scss-foundation/src/modules/typography/_fonts.scss
@@ -1,0 +1,7 @@
+@mixin set-font($family, $size: 16px, $weight: 400) {
+	font: {
+		family: $family;
+		size: $size;
+		weight: $weight;
+	};
+}
--- a/libs/scss-foundation/src/modules/typography/_fonts.scss
+++ b/libs/scss-foundation/src/modules/typography/_fonts.scss
@@ -1,11 +1,9 @@
 @use 'sass-true';
+@use 'sass-true/sass/throw';
 @use '../../modules/variables' as vars;
 
 @mixin define-font($family: null, $size: null, $weight: null) {
-	@if not $family and not $size and not $weight {
-		@include sass-true.error('At least one of the font properties (family, size, weight) must be provided.');
-	}
-	@else {
+	@if ($family and $size) or ($family and $weight) or ($size and $weight) {
 		$properties: (
 			'family': $family,
 			'size': $size,
@@ -17,5 +15,12 @@
 				font-#{$property}: $value;
 			}
 		}
+	}
+	@else {
+		@include throw.error(
+			$message: 'At least two of the font properties must be provided.',
+			$source: 'define-font($family, $size, $weight)',
+			$catch: vars.$testing
+		);
 	}
 }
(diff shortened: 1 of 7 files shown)


libs/scss-foundation/src/modules/typography/_fonts.scss
```diff
@@ -0,0 +1,13 @@
+@mixin define-font($family: null, $size: null, $weight: null) {
+	font: {
+		@if $family {
+			family: $family;
+		}
```
typhoon41, line 5: Even if we make all properties optional, at least one must be given, otherwise mixin doesn't have a point.
So...If all properties end up optional, we need to add validation when all are null to throw - and cover it in tests.

libs/scss-foundation/tests/typography/fonts.tests.scss
```diff
@@ -0,0 +1,75 @@
+@use 'sass-true/sass/true';
+@use '../../src/modules/typography/fonts';
+
+@include true.describe('define-font()') {
+
+	@include true.it('should set the font family') {
```
typhoon41, line 6: We don't need this many tests for mixin as simple as current one.
Remember our talks about first counting the number of tests needed? Number of branching + special cases.
We have 3 ifs + special case might be the bigger picture (when everything's given) = 4 tests and we have 6 atm...

libs/scss-foundation/src/modules/typography/_fonts.scss
typhoon41: As per our last discussion, we actually and probably want two out of three non-nullable properties.
Check needs to look like 
`@if not ($family and $size and $weight) {
    @include sass-true.error
}`

libs/scss-foundation/tests/typography/fonts.tests.scss
typhoon41: Validation tests are missing (that needs to also be adjusted - 2 out of 3), which will now increase number of tests :)

libs/scss-foundation/src/modules/typography/_fonts.scss
jovana-marceta: In the meantime, I  added throw.error to ensure that at least two font properties  are provided when using the define-font mixin. I also covered this error message in tests to verify correct behavior.

libs/scss-foundation/tests/typography/fonts.tests.scss
```diff
@@ -0,0 +1,57 @@
...
+			@include true.output() {
+				@include fonts.define-font($size: 18px, $weight: 700);
+			}
+			@include true.expect() {
+				font-size: 18px;
+				font-weight: 700;
+			}
+		}
+	}
+
```
typhoon41, line 44: Two tests cases are missing: Family & size + Family & weight.

typhoon41 (approved):

Cvetanovic (approved):

PR #234, opened 2024-08-26
BP-1225: Update stylelint depricated packages and rules

jovana-marceta:
BP-1225


libs/stylelint-config/index.js
```diff
@@ -118,7 +118,9 @@ module.exports = {
     "declaration-block-no-duplicate-custom-properties": true,
     "declaration-block-no-duplicate-properties": true,
     "declaration-block-no-shorthand-property-overrides": true,
-    "declaration-block-no-redundant-longhand-properties": true,
+    "declaration-block-no-redundant-longhand-properties": [
+      true, { "ignoreShorthands": [] }
```
jovana-marceta, line 122: The ignoreLonghands: [] option was introduced in Stylelint version 16.4.0. We might want to consider whether there are any scenarios where certain longhand properties should be ignored
https://stylelint.io/user-guide/rules/declaration-block-no-redundant-longhand-properties/

Mouvedia: Depending on the browsers you are targetting that would be `"text-decoration-thickness"` and/or `"background-size", "background-origin", "background-clip"`.

Mouvedia: Depending on the browsers you are targetting that would be `"text-decoration-thickness"` and/or `"background-size", "background-origin", "background-clip"`.

typhoon41: Hi @Mouvedia! Many thanks for all the efforts in making Stylelint great library and helping us here as well!
Keep up the good work!

As for our support, we do not care of IE or older versions of browsers but Safari happens to be the most fragile lately. I'd go with empty array unless you had experience with some of the mentioned ones in combination with Safari?

typhoon41 (changes requested):

libs/stylelint-config/index.js
Mouvedia: 👋🏻

> I'd go with empty array unless you had experience with some of the mentioned ones in combination with Safari?

check https://caniuse.com/?search=text-decoration-thickness and https://caniuse.com/background-img-opts

typhoon41: Still sounds like an empty array :) I guess there's only one way to find out...

Mouvedia: You might want to add `"transition-behavior"` retroactively.
i.e. it's not supported by the rule yet

libs/stylelint-config/package.json
```diff
@@ -15,18 +15,17 @@
...
     "postcss-scss": "^4.0.9",
     "stylelint-config-property-sort-order-smacss": "^10.0.0",
-    "@stylistic/stylelint-plugin": "^2.1.2",
+    "@stylistic/stylelint-plugin": "^3.0.1",
     "stylelint-use-nesting": "^5.1.1",
     "stylelint-declaration-block-no-ignored-properties": "^2.8.0",
     "stylelint-file-max-lines": "^1.0.0",
-    "stylelint-scss": "^6.1.0",
+    "stylelint-scss": "^6.5.1",
     "stylelint-group-selectors": "^1.0.10"
```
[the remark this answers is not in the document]
jovana-marceta: I tried to submit a pull request with the updates, but since I'm not a contributor to the project, I don't have the required permissions to commit to the repository. Fortunately, one of the existing contributors has created the PR, but it's currently pending approval from the original library author.

libs/stylelint-config/package.json
```diff
@@ -15,18 +15,17 @@
...
   },
   "peerDependencies": {
-    "stylelint": "^16.5.0"
+    "stylelint": "^16.8.2"
```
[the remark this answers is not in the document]
jovana-marceta: As usual, I reviewed the Stylelint releases to see what's been fixed and added since version 16.2.1 (the version before Tamara's update), including versions 16.5.0 and newer releases. I extended the selector-max-compound-selectors rule with ignoreSelectors and added the no-unknown-custom-media rule introduced in 16.4.0. All other changes were fixes. In the meantime, latest Stylelint version is released (16.9.0), I'll update to that as well."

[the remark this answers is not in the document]
jovana-marceta: For the @stylistic/stylelint-plugin, I reviewed the releases from version 2.1.0 and updated to 3.0.1 (which I confirmed is compatible with Stylelint 16.8.2). The only new rule is named-grid-areas-alignment for grid-template-areas, but since we don't use that feature, I thought we might not need it

typhoon41 (approved):

Cvetanovic (approved):

PR #237, opened 2024-09-27
BP-1194: As a user, I want the UI to be WCAG 2.1 compliant to ensure accessibility for all users

Cvetanovic:
https://enigmatry.atlassian.net/browse/BP-1247
https://enigmatry.atlassian.net/browse/BP-1248

### DIFF
--- a/libs/entry-components/dialog/dialogs/entry-dialog.component.html
+++ b/libs/entry-components/dialog/dialogs/entry-dialog.component.html
@@ -1,8 +1,10 @@
-<div class="entry-dialog">
+<div class="entry-dialog" role="dialog">
     <div class="dialog-header">
-        <h1 class="title" [ngClass]="{'without-close-icon': hideClose === true}" mat-dialog-title>{{ title }}</h1>
-        <button mat-icon-button *ngIf="!hideClose" type="button" class="close-button" (click)="cancel()">
-            <mat-icon>close</mat-icon>
+        <h1 class="title" tabindex="0" [ngClass]="{'without-close-icon': hideClose === true}" mat-dialog-title>{{ title
+            }}</h1>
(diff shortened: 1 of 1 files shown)


libs/entry-components/dialog/dialogs/entry-dialog.component.html
```diff
@@ -1,8 +1,10 @@
-<div class="entry-dialog">
+<div class="entry-dialog" role="dialog">
     <div class="dialog-header">
-        <h1 class="title" [ngClass]="{'without-close-icon': hideClose === true}" mat-dialog-title>{{ title }}</h1>
-        <button mat-icon-button *ngIf="!hideClose" type="button" class="close-button" (click)="cancel()">
-            <mat-icon>close</mat-icon>
+        <h1 class="title" tabindex="0" [ngClass]="{'without-close-icon': hideClose === true}" mat-dialog-title>{{ title
```
jovana-marceta, line 3: You can replace this class binding with a simpler and more readable syntax since we're only toggling one class. You can omit === true because hideClose will inherently evaluate to true or false based on its value due to JavaScript's truthy and falsy behavior:
`[class.without-close-icon]="hideClose"`

msljivic: I agree

msljivic, line 3: So title should be focusable?

jovana-marceta (approved):

libs/entry-components/dialog/dialogs/entry-dialog.component.html
Cvetanovic: no, I will remove that, screen readers will read title if we have role, etc, and with keyboard, titles generally don't require user interaction

msljivic (approved):

PR #240, opened 2024-10-30
BP-1293: Add theming support to new rich-text editor for basic redactor configuration

jovana-marceta:
https://enigmatry.atlassian.net/browse/BP-1293

Redactor will be used to replace CKEditor. The new rich text editor comes with some predefined CSS, and doesn't support our theming system including colors, fonts, spacings and other customizable options.

This issue introduces a mixin to override default values for the following properties enabled in the Basic Redactor configuration:

**First Iteration Includes:**
- text (paragraphs) font family, font size and color 
- list items -> same as text
- headings (h1-h4 is supported) font family
- blockquote border

**Further Considerations:**
- check if h5 & h6 should be covered
- checkboxes (on the official site there is an example with custom checkboxes to be checked)
- check elements that come with Standard and Custom Redactor configurations
- Dark/light theme support: Currently, Redactor defaults to the system theme, but this can be overridden with CSS. A better solution would involve adding a theme: 'light' option in the Redactor configuration file. Including this as a default in our Redactor repository might be beneficial since dark theme support is not yet available.

### DIFF
--- a/libs/entry-components/styles/_generator.scss
+++ b/libs/entry-components/styles/_generator.scss
@@ -8,6 +8,7 @@
 @use 'modules/components/toggle/generator' as toggle;
 @use 'modules/default-theme' as default;
 @use 'modules/vendors/angular-material/generator' as material-theme;
+@use 'modules/vendors/redactor/generator' as redactor;
 @use 'sass:map';
 @use 'partials/theming';
 
@@ -23,4 +24,5 @@
 	@include dialogs.generate-from($merged-theme);
 	@include checkboxes.generate-from($merged-theme);
 	@include toggle.generate-from($merged-theme);
+	@include redactor.generate-from($merged-theme);
 }
--- a/libs/entry-components/styles/modules/vendors/redactor/_generator.scss
+++ b/libs/entry-components/styles/modules/vendors/redactor/_generator.scss
@@ -1,0 +1,45 @@
+@use 'sass:map';
+
+$redactor-checkboxes-selector: '[data-rx-type = 'todoitem'] input:checked::before';
+
+@mixin generate-from($theme) {
+	$theming-fonts: map.get($theme, 'general', 'fonts');
+	$theming-colors: map.get($theme, 'general', 'colors');
+	$theming-checkboxes: map.get($theme, 'general', 'checkboxes');
+
+	.entry-redactor {
+		.rx-content {
+			@if $theming-colors {
+				p, li, h1, h2, h3, h4, h5, h6 {
+					color: map.get($theming-colors, 'font');
+				}
+	
+				blockquote {
+					border-left-color: map.get($theming-colors, 'primary');
+				}
+			}
+	
+			@if $theming-fonts {
+				p, li {
+					font-family: map.get($theming-fonts, 'body', 'family');
+					font-size: map.get($theming-fonts, 'body', 'size');
+				}
+			
+				h1, h2, h3, h4 {
+					font-family: map.get($theming-fonts, 'hero-titles', 'family');
+				}
+			
+				h5, h6 {
+					font-family: map.get($theming-fonts, 'titles', 'family');
+				}
+			}
+		}
+
+		@if $theming-checkboxes {
+			.rx-editor #{$redactor-checkboxes-selector} {
+				border-color: map.get($theming-checkboxes, 'background');
+				background-color: map.get($theming-checkboxes, 'background');
+			}
+		}
+	}
+}


libs/entry-components/styles/modules/vendors/redactor/_buttons-generator.scss
```diff
@@ -0,0 +1,22 @@
+@use 'sass:map';
+
+$material-button-background: #237BBB;
+
+@mixin generate-from($theme) {
+
+	$theming-buttons: map.get($theme, 'general', 'fonts', 'buttons');
+
+	.enigmatry-redactor-content button {
```
jovana-marceta, line 9: This is a custom class that I added in Redactor lib, named enigmatry instead of entry to follow convention for internal packages, to be able to override Redactor default styling.

libs/entry-components/styles/modules/vendors/redactor/_buttons-generator.scss
```diff
@@ -0,0 +1,22 @@
+@use 'sass:map';
+
+$material-button-background: #237BBB;
```
jovana-marceta, line 3: As agreed, if we don’t add specific values in our custom theme, it should inherit Angular Material's default values. However, this depends on the Redactor configuration on project level because there are two ways of adding it. If Redactor is passed within generated Formly components, it will have a Material container wrapper and will inherit values by default; otherwise, if we create custom redactor configuration, it will not inherit values so I added default values so now, both cases are covered.

libs/entry-components/styles/modules/vendors/redactor/_tables-generator.scss
```diff
@@ -0,0 +1,29 @@
+@use 'sass:map';
+@use '../../../../../scss-foundation/src/modules/lists/row-coloring' as list;
```
jovana-marceta, line 2: @typhoon41 @Cvetanovic  I used a mixin from scss-foundation here, but I’m concerned about what will happen if the scss-foundation package isn’t installed at the project level. Should I create a custom mixin specifically for Redactor row coloring to be safe? I also haven't found a way to avoid using relative paths

typhoon41 (changes requested):

libs/entry-components/styles/modules/vendors/redactor/_tables-generator.scss
```diff
@@ -0,0 +1,29 @@
...
+	.enigmatry-redactor-content table {
+		@if $theming-table {
+			tr {
+				@include list.odd-row-coloring(map.get($theming-table, 'odd'));
+				@include list.even-row-coloring(map.get($theming-table, 'even'));
+			}
+		}
+		@else {
+			tr {
+				@include list.even-row-coloring($material-odd-background);
```
Cvetanovic, line 18: if is this for even row then var name should be material-even-background

libs/entry-components/styles/partials/vendors/overrides/redactor/_general.scss
```diff
@@ -0,0 +1,38 @@
...
+		color: $material-text-color;
+	}
+
+	#{$redactor-todoitem-selector} {
+		input {
+			width: $material-checkbox-width;
+			height: $material-checkbox-height;
+		}
+
+		input::before, input:checked::before {
```
Cvetanovic, line 23: if this is related to checkbox and input, then it should be in the forms file, not in general

libs/entry-components/styles/modules/vendors/redactor/_tables-generator.scss
jovana-marceta: Now we get values from Material, so no more hardcoding

libs/entry-components/styles/partials/vendors/overrides/rich-text/_buttons.scss
```diff
@@ -0,0 +1,36 @@
...
+		border: none;
+		text-decoration: none;
+	}
+
+	a:hover {
+		background-color: mat.get-color-from-palette(mat.$grey-palette, 100);
+	}
+
+	button {
+		@include mat.elevation(4);
```
jovana-marceta, line 27: This is aa mixin provided by Angular Material to apply elevation styles (shadows) to an element. The number 4 represents the elevation level, which determines the intensity of the shadow so now we will have original dynamic values from Material

libs/entry-components/styles/partials/vendors/overrides/rich-text/_general.scss
```diff
@@ -0,0 +1,12 @@
+@use 'sass:map';
+@use '@angular/material' as mat;
+
+.enigmatry-redactor-content {
+	.rx-button-icon svg {
+		fill: rgb(0 0 0 / .54);
```
jovana-marceta, line 6: Angular Material’s built-in palettes don’t provide direct shades of black (rgb(0, 0, 0)) with specific opacities like .38 or .54. The existing palettes, such as $grey-palette, offer different shades of gray but don’t include true black with those exact transparency levels. Using rgba in this way allows us to achieve the exact colors and opacity levels without needing to rely on Angular Material’s palette system for custom black opacities.

libs/entry-components/styles/partials/vendors/overrides/rich-text/_forms.scss
```diff
@@ -0,0 +1,20 @@
+@use '@angular/material' as mat;
+
+$checkbox-width: 18px;
```
jovana-marceta, line 3: Since the Redactor component generates its own DOM elements and doesn’t directly use Angular Material components, it can’t directly inherit Angular Material's checkbox dimensions. Additionally, Angular Material doesn’t expose the checkbox width and height as SCSS variables, so to avoid any confusion, the most practical solution is to set custom SCSS variables to override redactor checkbox width and height that approximate Angular Material’s dimensions.
 image

libs/entry-components/styles/modules/vendors/rich-text/necessary-mix/_typography-generator.scss
```diff
@@ -0,0 +1,78 @@
...
+$hero-typography: map.get(mat.define-typography-config(), 'headline-1');
+$title-typography: map.get(mat.define-typography-config(), 'subtitle-1');
+$caption-typography: map.get(mat.define-typography-config(), 'caption');
+
+@mixin generate-from($theme) {
+	$theming-fonts: map.get($theme, 'general', 'fonts');
+
+	.enigmatry-redactor-content {
+		.rx-editor, .rx-content, .rx-content p {
+			@if $theming-fonts and map.has-key($theming-fonts, 'body') and map.get($theming-fonts, 'body', 'color') != null {
```
jovana-marceta, line 15: I attempted various approaches to avoid the null value and bypass the stylelint rule, but unfortunately, none of them were successful.

libs/entry-components/styles/modules/vendors/rich-text/necessary-mix/_tables-generator.scss
```diff
@@ -0,0 +1,56 @@
...
+		table {
+			@if $theming-table {
+				tr {
+					@include list.odd-row-coloring(map.get($theming-table, 'odd'));
+					@include list.even-row-coloring(map.get($theming-table, 'even'));
+				}
+			}
+			@else {
+				tr {
+					@include list.even-row-coloring(mat.get-color-from-palette(mat.$grey-palette, 100));
```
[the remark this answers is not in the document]
jovana-marceta: It matches the most original material color for even row

libs/entry-components/styles/modules/vendors/rich-text/necessary-mix/_buttons-generator.scss
```diff
@@ -0,0 +1,46 @@
...
+
+$default-material-button: map.get(mat.define-typography-config(), 'button');
+
+@mixin generate-from($theme) {
+	$theming-buttons: map.get($theme, 'general', 'fonts', 'buttons');
+
+	.enigmatry-redactor-content {
+		button, .rx-content button {
+			
+			@if (theming-buttons, 'colors', 'primary') {
```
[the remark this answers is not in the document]
typhoon41: That's fine. Short and sweat, but where did you find it? Nothing about that in official docs AFAIK?

typhoon41 (changes requested):

libs/entry-components/styles/modules/vendors/rich-text/necessary-mix/_buttons-generator.scss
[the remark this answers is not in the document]
jovana-marceta: The ability to pass multiple keys to map.get is a feature supported in Dart Sass, as described in the official Sass documentation. This functionality allows for concise access to deeply nested values in maps without chaining multiple map.get calls manually. It simplifies the process of working with complex nested data structures like our $theme map

This feature is documented in [the official Sass sass:map module](https://sass-lang.com/documentation/modules/map/)

Additionally, here are GitHub issuse tracking the feature:
https://github.com/sass/sass/issues/1739
https://github.com/sass/dart-sass/issues/1884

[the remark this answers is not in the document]
typhoon41: This is not about map.get keys. We're using that on various places. The question was about if condition joined with such map syntax?
IDK how this would even compile: @if ($theming-buttons, 'colors', 'primary')

Is the intention to map.get primary color and then compare it with unset (null) value?

PR #241, opened 2024-11-14
BP-1292: In order to see the errors in the form, change entry-search-filter component to display the form errors  

apa-enigmatry:
https://enigmatry.atlassian.net/browse/BP-1292

### DIFF
--- a/apps/demo-app/src/app/examples/search-filter/search-filter/search-filter-example.component.ts
+++ b/apps/demo-app/src/app/examples/search-filter/search-filter/search-filter-example.component.ts
@@ -1,8 +1,9 @@
-import { Component } from '@angular/core';
+import { Component, ViewChild } from '@angular/core';
 import { UsersService } from './users.service';
 import {
   AutocompleteSearchFilter,
   DateTimeSearchFilter,
+  EntrySearchFilterComponent,
   SearchFilterBase,
   SearchFilterParams,
   SelectOption,
@@ -19,6 +20,8 @@ import { of } from 'rxjs';
   styleUrls: ['./search-filter-example.component.scss']
 })
 export class SearchFilterExampleComponent {
+  @ViewChild(EntrySearchFilterComponent) entrySearchFilterComponent: EntrySearchFilterComponent;
+
   users: Array<User>;
   displayedColumns: string[] = ['name', 'email', 'dateOfBirth', 'occupation', 'country'];
   filters = [];
@@ -29,7 +32,30 @@ export class SearchFilterExampleComponent {
   }
 
   searchFilterChange(searchParams: SearchFilterParams) {
-    this.fetchUsers(searchParams);
+    const errors: { [key: string]: string[] } = {};
+
+    // Simulate a backend validation error for dateOfBirth
+    if (searchParams.dateOfBirth && this.isInvalidDate(searchParams.dateOfBirth)) {
+      errors.dateOfBirth = ["The date cannot be in the future."];
+    }
+
+    if (Object.keys(errors).length > 0) {
+      this.setServerErrors(errors);
+    } else {
+      this.fetchUsers(searchParams);
+    }
+  }
+
+  private isInvalidDate(date: string): boolean {
+    const selectedDate = new Date(date);
+    const maxDate = new Date();
+    return selectedDate > maxDate;
+  }
+
+  setServerErrors(errors: { [key: string]: string[] }) {
+    if (this.entrySearchFilterComponent) {
+      this.entrySearchFilterComponent.setServerErrors(errors);
+    }
   }
 
   private fetchUsers(searchParams: SearchFilterParams = {}): void {
--- a/apps/demo-app/src/app/examples/search-filter/search-filter/search-filter-example.component.ts
+++ b/apps/demo-app/src/app/examples/search-filter/search-filter/search-filter-example.component.ts
@@ -10,9 +10,10 @@ import {
   SelectSearchFilter,
   TextSearchFilter,
 } from '@enigmatry/entry-components/search-filter';
-import { map } from 'rxjs/operators';
+import { map, tap } from 'rxjs/operators';
 import { Country, Occupation, User } from './users';
-import { of } from 'rxjs';
+import { Observable, of } from 'rxjs';
+import { IValidationProblemDetails, setServerSideValidationErrors } from '@enigmatry/entry-components';
 
 @Component({
   selector: 'app-search-filter-example',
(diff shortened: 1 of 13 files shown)


libs/entry-components/search-filter/entry-search-filter.module.ts
```diff
@@ -14,6 +14,7 @@ import { DateTimeSearchFilterComponent } from './date-time/date-time-search-filt
 import { MatDatepickerModule } from '@angular/material/datepicker';
 import { MatFormFieldModule } from '@angular/material/form-field';
 import { DateSearchFilterComponent } from './date/date-search-filter.component';
+import { EntryValidationModule } from '../validation/entry-validation.module';
```
msljivic, line 17: Should be ```import { EntryValidationModule } from '@enigmatry/entry-components/validation';```

msljivic: Paths inside module can be referenced with local path (./)
Paths outside from other modules, must be referenced with @enigmatry/entry-components/...

libs/entry-components/styles/partials/core/components/search-filter/_general.scss
```diff
@@ -1,9 +1,6 @@
 @use 'scss-foundation/src/modules/display/items';
 
 .search-form-container {
+	padding-bottom: 10px;
 	@include items.align-center();
```
jovana-marceta, line 5: Place @include directive at the beginning of a code block, before any other declarations for better readability and consistency

libs/entry-components/search-filter/date-time/date-time-search-filter.component.html
```diff
@@ -1,6 +1,7 @@
-<mat-form-field [formGroup]="form" subscriptSizing="dynamic">
+<mat-form-field class="date-time-search-field" [formGroup]="form" subscriptSizing="dynamic">
     <mat-label [attr.for]="searchFilter.key">{{searchFilter.label}}</mat-label>
     <input matInput [formControlName]="searchFilter.key" [matDatepicker]="picker" [id]="searchFilter.key">
     <mat-datepicker-toggle matIconSuffix [for]="picker"></mat-datepicker-toggle>
     <mat-datepicker #picker></mat-datepicker>
+    <mat-error entryDisplayControlValidation [control]="form.get(searchFilter.key)"></mat-error>
```
zsrdjan, line 6: @msljivic Just a minor comment, but what do you think about having a control that will be deriving from control value accessor? This would allow us to use [formControlName]="searchFilter.key" syntax. 
E.g.
`<entry-form-field-error [formControlName]="searchFilter.key"/>`

libs/entry-components/search-filter/entry-search-filter.component.ts
```diff
@@ -69,5 +72,13 @@ export class EntrySearchFilterComponent implements OnInit {
   asDateSearchFilter<T>(searchFilter: SearchFilterBase<T>): DateSearchFilter<T> {
     return searchFilter as DateSearchFilter<T>;
   }
-}
 
+  setServerErrors(errors: { [key: string]: string[] }) {
```
msljivic, line 76: There is a helper method that does this (this code can be removed) See: https://github.com/enigmatry/entry-angular-building-blocks/blob/9df5d90230a79adca8f5d8d79771b4f12dd44cdc/libs/entry-components/validation/entry-validation.ts#L29

(method that maps received server side validation errors to the form and its fields.)

apps/demo-app/src/app/examples/search-filter/search-filter/search-filter-example.component.ts
```diff
@@ -29,7 +32,30 @@ export class SearchFilterExampleComponent {
   }
 
   searchFilterChange(searchParams: SearchFilterParams) {
-    this.fetchUsers(searchParams);
+    const errors: { [key: string]: string[] } = {};
```
msljivic, line 35: This example is a bit misleading. We want to show how to display server side validation errors in search filters. It should look like something like this:

```ts
import { setServerSideValidationErrors } from '@enigmatry/entry-components/validation'
....
....
this.fetchUsers(searchParams)
      .subscribe({
        error: (error: IValidationProblemDetails) => {
          /** Applies received server side validation errors to the search filters form */
          setServerSideValidationErrors(error, this.entrySearchFilterComponent.searchFilterForm);
        }
      });
```
I would move mocked server side validation response in service and return it from there.

apps/demo-app/src/app/examples/search-filter/search-filter/search-filter-example.component.ts
```diff
@@ -19,6 +20,8 @@ import { of } from 'rxjs';
   styleUrls: ['./search-filter-example.component.scss']
 })
 export class SearchFilterExampleComponent {
+  @ViewChild(EntrySearchFilterComponent) entrySearchFilterComponent: EntrySearchFilterComponent;
```
msljivic, line 23: Whenever possible use { static: true } on the @ViewChild query. In this case we should use it.

libs/entry-components/search-filter/date-time/date-time-search-filter.component.html
msljivic: It would not be a appropriate use of CVA. Directive `entryDisplayControlValidation` seems fine, it's used only to display validation error so it's applied to <mat-error> element.

If we would implement CVA, implemented methods (writeValue, registerOnChange....) would have no purpose.

msljivic: Usually errors are displayed next to input. So control that wraps input and validation display (e.g. entry-input) would implement CVA

zsrdjan: My motivation comes from writing more condensed, readable, maintainable code. 

This is way more readabable and thus more difficult to make a mistake:
`<entry-form-field-error [formControlName]="searchFilter.key"/>`
compared to this:
`<mat-error entryDisplayControlValidation [control]="form.get(searchFilter.key)"></mat-error>`

Is there another way to achieve this condensity? I do favor readability over "correctness" of implementation. I'm not even sure that I agree that this is "correct".

zsrdjan: @msljivic what do you think about this as an alternative:
`<entry-error-display [control]="form.get('field')"></entry-error-display>`

And component:
```
@Component({
  selector: 'entry-error-display',
  template: `
    <mat-error *ngIf="control.invalid && (control.dirty || control.touched)">
      <!-- Add other error messages based on control.errors -->
    </div>
  `
})
export class EntryErrorDisplayComponent {
  @Input() control: AbstractControl;
}
```

apps/demo-app/src/app/examples/search-filter/search-filter/search-filter-example.component.ts
apa-enigmatry: I will move it to the service

msljivic (approved):

jovana-marceta (approved):

libs/entry-components/search-filter/date-time/date-time-search-filter.component.html
msljivic: Component alternative is ok. We can do it that way.

PR #246, opened 2024-12-17
BP-1372: Bump all Stylelint package versions (and adjust related rule changes)

Cvetanovic:
https://enigmatry.atlassian.net/browse/BP-1372


package.json
```diff
@@ -70,7 +70,7 @@
     "typedoc-plugin-markdown": "^4.2.10",
     "typescript": "5.5.4",
     "typescript-eslint": "^7.18.0",
-    "stylelint": "^16.10.0",
+    "stylelint": "^16.12.0",
```
jovana-marceta, line 73: Did you also update stylelint package from stylelint-config package.json file? We need to update on both places

jovana-marceta (approved):

libs/stylelint-config/package.json
```diff
@@ -23,7 +23,7 @@
     "stylelint-group-selectors": "^1.0.10"
   },
   "peerDependencies": {
-    "stylelint": "^16.10.0"
+    "stylelint": "^16.12.0"
```
jovana-marceta, line 26: Should we consider adding the selector-pseudo-class-allowed-list rule to our styling configuration? Since this rule has been extended in the newest version of our stylint.

typhoon41: ...And even if we don't want it, we still miss it in our config, so - yes.

typhoon41:
stylelint-scss package: function-color-channel was added in 6.8.0 but is missing in our config
stylelint-use-nesting package: csstools/use-nesting not configured at all in our config
stylelint package: media-query-no-invalid rule missing...please check if any other rule is missing (compare rule list from website and our config)

typhoon41 (changes requested):

package.json
```diff
@@ -70,14 +70,14 @@
...
     "postcss-scss": "^4.0.9",
     "stylelint-config-property-sort-order-smacss": "^10.0.0",
     "@stylistic/stylelint-plugin": "^3.1.1",
     "stylelint-use-nesting": "^6.0.0",
```
typhoon41, line 77: There is no version 6.0?

https://github.com/csstools/stylelint-use-nesting

libs/stylelint-config/index.js
```diff
@@ -248,9 +333,7 @@ module.exports = {
     "plugin/declaration-block-no-ignored-properties": true,
     "plugin/stylelint-group-selectors": true,
     //"plugin/stylelint-selector-no-empty": true, => uncomment this line once this library is again compatible with stylelint v15+
-    // "csstools/use-nesting": ["always", => uncomment this line once this library is again compatible with stylelint v16+
-    //   { "syntax": "scss" }
-    // ],
+    "csstools/use-nesting": ["always", { "syntax": "scss" }],
```
typhoon41, line 336: How do we know this works on Stylelint v16+? The latest version is 5.1.1 and was release almost a year ago (March 6th 2024)...

libs/stylelint-config/index.js
```diff
@@ -184,35 +235,68 @@ module.exports = {
...
       { "ignoreAtRules": ["use"] }
     ],
     "scss/at-use-no-redundant-alias": true,
+    "scss/at-use-no-unnamespaced": null,
     "scss/block-no-redundant-nesting": true,
+    "scss/declaration-nested-properties": null,
+    "scss/declaration-nested-properties-no-divided-groups": null,
+    "scss/dimension-no-non-numeric-values": null,
     "scss/dollar-variable-colon-space-after": "always",
     "scss/dollar-variable-colon-space-before": "never",
```
typhoon41, line 287: This file is not sorted anymore...We want it sorted so we can easily find rule we need.

Please do not sort manually, you can use Notepad++ or any other tool.

libs/stylelint-config/index.js
```diff
@@ -184,35 +235,68 @@ module.exports = {
...
     "scss/at-if-no-null": true,
+    "scss/map-keys-quotes": null,
     "scss/at-mixin-argumentless-call-parentheses": "always",
     "scss/at-mixin-parentheses-space-before": "never",
     "scss/at-root-no-redundant": true,
+    "scss/at-rule-conditional-no-parentheses": null,
     "scss/at-rule-no-unknown": [
       true,
       { "ignoreAtRules": ["use"] }
     ],
```
typhoon41, line 279: I think following rules needs to be reconsidered:

at-rule-empty-line-before
declaration-property-value-keyword-no-deprecated
keyframe-selector-notation (percentage-unless-within-keyword-only-block option)
selector-max-class (6 as a max value?)
selector-max-specificity (0,5,0?)
scss/load-no-partial-leading-underscore
scss/at-import-no-partial-leading-underscore
scss/at-import-partial-extension (never)
scss/at-rule-conditional-no-parentheses
scss/at-use-no-unnamespaced

typhoon41 (changes requested):

package.json
Cvetanovic: yes you are right for that link, but here that version exists...interesting
https://www.npmjs.com/package/stylelint-use-nesting

typhoon41: let's stick to 5.1.1 until they document it

libs/stylelint-config/package.json
```diff
@@ -15,15 +15,15 @@
...
     "stylelint-use-nesting": "^6.0.0",
     "stylelint-declaration-block-no-ignored-properties": "^2.8.0",
     "stylelint-file-max-lines": "^1.0.0",
-    "stylelint-scss": "^6.10.0",
+    "stylelint-scss": "^6.11.1",
     "stylelint-group-selectors": "^1.0.10"
   },
   "peerDependencies": {
-    "stylelint": "^16.10.0"
+    "stylelint": "^16.15.0"
```
typhoon41, line 26: We should remove peer dependencies. We don't want anybody to use any other version we do not support.

package.json
```diff
@@ -68,12 +68,12 @@
     "postcss-scss": "^4.0.9",
     "sass": "1.83.4",
     "sass-true": "8.1.0",
-    "stylelint": "^16.13.2",
+    "stylelint": "^16.15.0",
```
typhoon41, line 71: 16.16 is the latest for 11 days now.

libs/stylelint-config/index.js
```diff
@@ -5,12 +5,12 @@ module.exports = {
   extends: "stylelint-config-property-sort-order-smacss",
   plugins: [
     "@stylistic/stylelint-plugin",
-    "stylelint-use-nesting",
     "stylelint-declaration-block-no-ignored-properties",
     "stylelint-file-max-lines",
-    "stylelint-scss",
     "stylelint-group-selectors",
     // "stylelint-selector-no-empty",  => uncomment this line once this library is again compatible with stylelint v16+
```
typhoon41, line 11: You can safely delete this comment. Seems there's no activity anymore in the package.

typhoon41 (changes requested):

typhoon41 (approved):

PR #247, opened 2024-12-30
Features/bp 1367 eslint updates

typhoon41:
https://enigmatry.atlassian.net/browse/BP-1367

### DIFF
--- a/libs/eslint-config/index.js
+++ b/libs/eslint-config/index.js
@@ -12,6 +12,9 @@ import arrowFunctions from "eslint-plugin-prefer-arrow-functions";
 export default tseslint.config(
   ...angular.configs.tsRecommended,
   {
+    ignores: ["src/**/generated/", "src/polyfills.ts", "src/@enigmatry/**", "**/api-reference.ts"]
+  },
+  {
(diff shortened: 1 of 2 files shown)


TamaraMarr (approved):
Generally, I don't have any specific feedback for the introduced changes, they seem fine, however, I do have to wonder if it is the best idea to keep maintaining this amount of rules. Maybe I don't have the necessary insight because I am joining in when everything is already set up in a specific way, but to me it seems that using recommended config and just adding specific adjustments to the few rules that prove to be bothersome during development could be a good approach.

typhoon41:
> Generally, I don't have any specific feedback for the introduced changes, they seem fine, however, I do have to wonder if it is the best idea to keep maintaining this amount of rules. Maybe I don't have the necessary insight because I am joining in when everything is already set up in a specific way, but to me it seems that using recommended config and just adding specific adjustments to the few rules that prove to be bothersome during development could be a good approach.

We'll definitely have a team call on this topic in hopefully, near future. We didn't do that in beginnings because back then subset of recommended rules were like 20% of all, which was too few to make a difference especially because they were missing stylistic rules.

PR #252, opened 2025-04-03
Made "None" option configurable in SelectSearchFilter

laslorajda:
BP-1455


libs/entry-components/search-filter/select/select-search-filter.model.ts
```diff
@@ -18,11 +18,13 @@ export class SelectSearchFilter<T> extends SearchFilterBase<T> {
    * If it is set to false, 'none selected' option becomes available as a first option.
    * */
   multiSelect = true;
+  allowNoneSelection = true;
```
msljivic, line 21: Maybe rename to `showNoneOption `

msljivic (approved):

PR #253, opened 2025-05-07
BP-1467: Fix Failing SCSS Unit Tests

Cvetanovic:
https://enigmatry.atlassian.net/browse/BP-1467

### DIFF
--- a/libs/scss-foundation/tests/layout/grid-core.tests.scss
+++ b/libs/scss-foundation/tests/layout/grid-core.tests.scss
@@ -79,14 +79,12 @@ $custom-breakpoints: () !default;
 			}
 
 			@include true.expect() {
-				/* stylelint-disable-next-line media-query-no-invalid */
 				@media only screen and (width >= 540px) {
 					.row.reverse-row-sm {
 						flex-direction: row-reverse;
 					}
 				}
 
-				/* stylelint-disable-next-line media-query-no-invalid */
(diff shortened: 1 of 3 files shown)


libs/scss-foundation/tests/layout/grid-core.tests.scss
```diff
@@ -83,29 +79,33 @@ $custom-breakpoints: () !default;
 			}
 
 			@include true.expect() {
+				/* stylelint-disable-next-line media-query-no-invalid */
```
typhoon41, line 82: Let's put all of these new comments to line 1 instead. These are tests (we know what we are doing), so let's do the way we did before and keep it small and simple.
Don't forget we still have all the rules turned on in function itself so violation of any of those rules will be visible there.

libs/scss-foundation/tests/layout/grid.tests.scss
```diff
@@ -133,778 +127,793 @@ vars.$testing: true;
...
 						flex-direction: row-reverse;
 					}
 				}
-					
+
 				.row.reverse-row-under-sm {
 					flex-direction: row-reverse;
-					
+
+					/* stylelint-disable-next-line media-query-no-invalid */
```
typhoon41, line 175: The same as in other file.

typhoon41 (approved):

PR #286, opened 2026-07-15
WSR-5653: Add opt-in option grouping to select and autocomplete components

stefanr89:
https://enigmatry.atlassian.net/browse/WSR-5653

Adds opt-in mat-optgroup support so select/autocomplete options can be rendered under group headers, without changing behavior for existing ungrouped consumers.

entry-form:
- SelectConfiguration/sortOptions gain an optional groupProperty, used to pre-sort and bucket options by group before Formly's own select/autocomplete rendering groups them.
- The custom autocomplete field type (select's stock Formly type already supported groups) now renders mat-optgroup, with the search filter and value/label lookups updated to recurse into grouped options.
- Demo: grouped single/multi select and autocomplete examples added, plus corresponding C# builder-API doc snippets.

entry-components/search-filter:
- SelectOption gains an optional group name; a new GroupSelectOptionsPipe buckets flat option lists into groups (pass-through when unused).
- select-search-filter and autocomplete-search-filter now render mat-optgroup when options are grouped.
- Demo: existing occupation (select) and country (autocomplete) filters updated to show grouping on real data.
- README updated with a grouping usage example.

Verified via ng build (full template type-checking), eslint, and Playwright-driven browser testing of every grouped/ungrouped select/autocomplete combination in both libraries - zero console errors, ungrouped paths confirmed byte-identical to prior behavior.

### DIFF
--- a/apps/demo-app/src/app/examples/search-filter/search-filter/search-filter-example.component.ts
+++ b/apps/demo-app/src/app/examples/search-filter/search-filter/search-filter-example.component.ts
@@ -68,10 +68,7 @@ export class SearchFilterExampleComponent {
         multiSelect: true,
         options: Object.values(Occupation)
           .filter(value => typeof value === 'number')
-          .map((value: number) => new SelectOption(
-            value,
-            Occupation[value].replace(/^[a-z]/u, x => x.toUpperCase()),
-            this.occupationGroup(value)))
+          .map((value: number) => new SelectOption(value, Occupation[value].replace(/^[a-z]/u, x => x.toUpperCase())))
       }),
       new SelectSearchFilter({
         key: 'username',
@@ -107,16 +104,6 @@ export class SearchFilterExampleComponent {
     ];
   }
 
-  private readonly occupationGroups: Partial<Record<Occupation, string>> = {
-    [Occupation.electrician]: 'Skilled trades',
-    [Occupation.plumber]: 'Skilled trades',
-    [Occupation.doctor]: 'Professional',
-    [Occupation.teacher]: 'Professional',
-    [Occupation.painter]: 'Creative',
-    [Occupation.baker]: 'Creative',
-    [Occupation.soldier]: 'Public service'
-  };
-
   private readonly countryContinents: Partial<Record<Country, string>> = {
     [Country.unitedStates]: 'Americas',
     [Country.canada]: 'Americas',
@@ -131,8 +118,6 @@ export class SearchFilterExampleComponent {
     [Country.australia]: 'Oceania',
     [Country.southAfrica]: 'Africa'
   };
-
-  private occupationGroup = (occupation: Occupation): string => this.occupationGroups[occupation] ?? 'Other';
 
   private countryContinent = (country: Country): string => this.countryContinents[country] ?? 'Europe';
 
--- a/apps/demo-app/src/app/examples/search-filter/search-filter/search-filter-example.component.ts
+++ b/apps/demo-app/src/app/examples/search-filter/search-filter/search-filter-example.component.ts
@@ -68,7 +68,10 @@ export class SearchFilterExampleComponent {
         multiSelect: true,
         options: Object.values(Occupation)
           .filter(value => typeof value === 'number')
-          .map((value: number) => new SelectOption(value, Occupation[value].replace(/^[a-z]/u, x => x.toUpperCase())))
+          .map((value: number) => new SelectOption(
+            value,
+            Occupation[value].replace(/^[a-z]/u, x => x.toUpperCase()),
+            this.occupationGroup(value)))
       }),
       new SelectSearchFilter({
         key: 'username',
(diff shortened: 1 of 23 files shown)


libs/entry-components/search-filter/group-select-options.pipe.ts
```diff
@@ -0,0 +1,38 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { SelectOption } from './select-option.model';
+
+export interface SelectOptionGroup {
```
mrle: I would change it to be generic type (`SelectOptionGroup<T>`) to match inner `SelectOption<T>`.
Also, would it be better if `label` (btw. maybe better name would be `groupLabel`) become nullable, in which case return type of the pipe could be `Array<SelectOptionGroup<any>>`. To elaborate a bit:

- In case when control contains un-grouped options, the pipe returns just a single group (`SelectOptionGroup`) with no name.
- In case when control contains grouped options, the pipe would return multiple groups.

This way the templates and the pipe could be made simplier.

mrle: Maybe a helper method in `SelectOptionGroup` named `HasGroup()` could be of help in templates too!?

mrle: Also, make this standalone file like `select-option.model.ts` is.

libs/entry-form/autocomplete/select-configuration.interface.ts
```diff
@@ -2,11 +2,38 @@ export interface SelectOption {
   value?: any;
   label: string;
   disabled?: boolean;
+  group?: SelectOption[];
```
mrle: Not sure if we should support more then 1 level of grouping?

Also, i was wondering, could we apply the same approach like you did for `search-filter` comeponent, to add `groupLabel?` instead of `group?`? We could apply similar filter like `groupSelectOptions` too ...

libs/entry-form/autocomplete/select-configuration.interface.ts
```diff
@@ -2,11 +2,38 @@ export interface SelectOption {
...
+    }
+  }
+  return undefined;
+};
+
+// Recursively looks up an option by label (case-insensitive), descending into `group` children.
+export const findOptionByLabel = (options: SelectOption[], label: string): SelectOption | undefined => {
+  const lowerLabel = label.toLowerCase();
+  for (const option of options) {
+    const found = option.group ?
```
mrle: Ternar operator (?:) formatting! Not sure, but to me it is more readable if ternar operator symbols begin at new line like:

```ts
const found = option.group
      ?  findOptionByLabel(option.group, label)
      :  option.label.toLowerCase() === lowerLabel ? option : undefined;
```

libs/entry-form/autocomplete/filter-with-autocomplete.pipe.ts
```diff
@@ -11,6 +11,10 @@ export class FilterWithAutocompletePipe implements PipeTransform {
       return options;
     }
     const labelStartsWith = filterWith.toLowerCase();
-    return options.filter(option => option.label.toLowerCase().includes(labelStartsWith));
+    return options
+      .map(option => option.group ?
```
mrle: Cauls some of `label.toLowerCase().includes(x)` expressions become a helper method of `SelectOption`?

Also check out my "Ternar operator (?:) formatting!" comment ;)

apps/demo-app/src/app/examples/search-filter/search-filter/search-filter-example.component.ts
```diff
@@ -104,6 +107,35 @@ export class SearchFilterExampleComponent {
     ];
   }
 
+  private readonly occupationGroups: Partial<Record<Occupation, string>> = {
```
mrle: Grouping both lists now removes examples without groups, which is basic usage.
These changes also apply to search-filter examples but without eny explanation in readme.

libs/entry-form/autocomplete/flatten-autocomplete-options.pipe.ts
```diff
@@ -0,0 +1,18 @@
...
+// label as groupName, so the rest of this module never has to deal with nesting.
+@Pipe({
+    name: 'flattenAutocompleteOptions',
+    standalone: false
+})
+export class FlattenAutocompleteOptionsPipe implements PipeTransform {
+  transform = (options: FormlySelectOption[] | null | undefined): SelectOption[] =>
+    (options ?? []).flatMap(option => option.group ?
+      option.group.map(child => ({ value: child.value, label: child.label, disabled: child.disabled, groupName: option.label })) :
+      [{ value: option.value, label: option.label, disabled: option.disabled }]);
```
msljivic: Optimize, we are creating arrays with single option just to flat map them after

msljivic (approved):

PR #288, opened 2026-08-07
BP-1582 - Upgrade to Angular 22

jovanasta:
# BP-1582 — Upgrade Angular to v22

Upgrades the workspace from Angular 21.2 → 22.1, Material/CDK 21.2 → 22.1, and
TypeScript 5.9 → 6.0. `angular-eslint` goes 21.3.1 → 22.1.0 because its major
peer-depends on `@angular/cli >= 22 < 23`; no other lint tooling is touched (see
*Deliberately not in this PR*).

Removed as dead weight: `@angular/animations` (Material 22 animates with native CSS),
`@angular/platform-browser-dynamic`, `jest-environment-node-single-context`.

## Consumer-visible breaking changes

Documented in a new `## Migrating to 22.x` section in `libs/entry-components/README.md`,
which `libs/entry-form/README.md` links to. There is no `CHANGELOG.md` in this repo, so
the README is the only channel consumers have.

1. **`@angular/animations` is no longer a peer dependency** — consumers must drop
   `provideAnimations()`.
2. **The router's `paramsInheritanceStrategy` default changed to `'always'`.** This
   weakens `entryPermissionGuard`: a child route with no `data.permissions` of its own now
   inherits its parent's instead of being denied. The demo app pins `'emptyOnly'`, the
   permissions README explains it, and the guard now logs a dev-mode warning when the
   permissions it checked came from an ancestor.

   The guard still reads resolved `route.data` rather than `routeConfig.data`, because the
   latter is not equivalent to pre-22 behaviour: under `'emptyOnly'` an empty-path or
   component-less child *does* inherit, so routes that legitimately inherit today would
   start being denied.
3. **`ChangeDetectionStrategy.OnPush` is now the framework default** — this reaches every
   consumer component that doesn't declare a strategy, not just ours.
4. **TypeScript 6 rejects `baseUrl`** (TS5101). Anyone relying on it for workspace-absolute
   imports needs explicit `paths` entries.

## Change detection

The PR accepts the new OnPush default rather than opting components back out.

- **Demo app** — eight components moved to signals: `example-viewer`, `spinner-example`,
  `search-filter-example`, both dialog examples (`confirm`, `custom`), and all three
  validation examples.
- **Libraries** — three components needed `ChangeDetectorRef`, because they mutate
  template-bound state from outside any Angular event:
  - `EntryFormErrorsComponent` — `setServerSideValidationErrors` mutates the form in
    place, so the bound reference never changes and nothing marks the view. Also fixes a
    pre-existing bug for consumers who already used OnPush.
  - `EntryTimePickerComponent` — `update()` and `to24HourClock()` are invoked imperatively
    by the parent, which marks the *parent's* view, not this child's.
  - `EntryFileInputComponent` — `writeValue()`, `setDisabledState()` and `clear()` are
    called by the forms API, never by a template event. Pre-existing latent bug; this
    component has been explicitly OnPush all along.

`EntryDateTimePickerComponent` is deliberately unchanged. A `markForCheck()` was added to
its `valueChanges` subscription during review and then reverted after testing showed it did
nothing: the visible input is bound `[formControl]`, so its value is written to the DOM by
the ControlValueAccessor with change detection out of the path, and `setValue()` emits
`statusChanges` too, where a sibling subscription already marks the view.

Other library changes: `SpinnerOverlayContainer` loses its constructor (CDK's
`OverlayContainer` moved to `inject()` internally).

## Toolchain and CI

- Node is pinned to `^24.15.0` via `UseNode@1` in a new `pipeline-templates/install-node.yml`,
  shared by both pipelines so the pin and its rationale live in one place.
- `engines.node` added in three places, which the repo previously had nowhere: `^24.15.0`
  on the workspace root so local dev matches CI, and the full Angular 22 range
  (`^22.22.3 || ^24.15.0 || >=26.0.0`) on both published packages, since consumers
  shouldn't inherit our choice of LTS line. Verified it survives ng-packagr into `dist`.
- `@types/node` held on the Node 24 line (`^24.13.3`) to match the runtime. Node 26 isn't
  LTS until October.
- `baseUrl` removed from the root `tsconfig.json`. `entry-codegen` emits workspace-absolute
  imports into `apps/demo-app/**/generated/`, so an `apps/*` mapping replaces it — and
  because every lib tsconfig inherits it through `extends`, `eslint.config.js` now bans
  `apps/**` imports from `libs/**`. The one non-generated file that used the mapping now
  imports relatively.

## Also included

- API documentation regenerated; agent instructions refreshed for v22.
- Demo cleanups from review: `reset()` extracted out of three example templates, Formly
  field groups tracked by `field.id` (was `track field.name`, which was `undefined` on
  every item and logged NG0955), underscore prefixes dropped from private members in the
  two library components this PR touches.

## Verification

There is no Angular unit test suite in this workspace, so this is lint-checked,
build-checked and exercised in a browser — not unit tested.

- `npm run lint` clean; `@enigmatry/entry-components`, `@enigmatry/entry-form` and the demo
  app all build; 50 SCSS tests pass.
- Demo app driven through every route. Each async path confirmed to repaint: spinner across
  its 3s timer, dialog result, table `resource()` load, search filter, date pickers, Formly
  form, and the validation examples' submit/reset. No NG0955 and no console errors; the only
  warnings are pre-existing and unrelated (Formly's `minlength` deprecation, the docs
  markdown sanitiser).
- The two library change-detection fixes were tested in both directions — removing the call
  and confirming the UI breaks — rather than assumed. That is how the date-time-picker one
  was caught as inert.


apps/demo-app/src/app/examples/validation/formly/complex-formly-form-validation/complex-formly-form-validation-example.component.html
```diff
@@ -5,11 +5,11 @@
     <entry-form-errors [form]="form"></entry-form-errors>
     <div class="action-buttons">
         <button mat-button entry-submit-button type="submit">Submit</button>
-        <button mat-button entry-cancel-button type="reset" (click)="form.reset(); validationResult = undefined;">Reset</button>
+        <button mat-button entry-cancel-button type="reset" (click)="form.reset(); validationResult.set(undefined);">Reset</button>
```
StojanovicMilos: why did we ever inline this in html? Just make a reset method that does this in ts. I know you didn't write the initial code, but let's apply boyscout rule

apps/demo-app/src/app/examples/validation/formly/complex-formly-form-validation/field-set.component.ts
```diff
@@ -6,7 +6,7 @@ import { FieldType } from '@ngx-formly/core';
     template: `@if (field.fieldGroup; as fieldGroup) {
                 <fieldset class="validation-field-group">
                     <legend class="validation-group-legend">{{props.label}}</legend>
-                    @for (field of fieldGroup; track field.name) {
+                    @for (field of fieldGroup; track $index) {
```
StojanovicMilos: was field.name not unique? Why is it replaced with $index?

apps/demo-app/src/app/examples/validation/formly/complex-formly-form-validation/repeat-name.component.ts
```diff
@@ -6,7 +6,7 @@ import { FieldArrayType } from '@ngx-formly/core';
     template: `
     <div class="validation-children-names">
       <label class="description">{{props.label}}:</label>
-      @for (field of field.fieldGroup; track field.name; let i = $index) {
+      @for (field of field.fieldGroup; track $index; let i = $index) {
```
StojanovicMilos: was field.name not unique? Why is it replaced with $index?

apps/demo-app/src/app/examples/validation/formly/formly-form-validation/formly-form-validation-example.component.html
```diff
@@ -6,11 +6,11 @@
     
     <div class="action-buttons">
         <button mat-button entry-submit-button type="submit">Submit</button>
-        <button mat-button entry-cancel-button type="reset" (click)="form.reset(); validationResult = undefined;">Reset</button>
+        <button mat-button entry-cancel-button type="reset" (click)="form.reset(); validationResult.set(undefined);">Reset</button>
```
StojanovicMilos: .

apps/demo-app/src/app/examples/validation/reactive-form/reactive-form-validation-example.component.html
```diff
@@ -17,11 +17,11 @@
 
     <div class="action-buttons">
         <button mat-button entry-submit-button type="submit">Submit</button>
-        <button mat-button entry-cancel-button type="reset" (click)="form.reset(); validationResult = undefined;">Reset</button>
+        <button mat-button entry-cancel-button type="reset" (click)="form.reset(); validationResult.set(undefined);">Reset</button>
```
StojanovicMilos: .

apps/demo-app/src/app/app-routing.module.ts
```diff
@@ -57,7 +57,10 @@ const routes: Routes = [
 ];
 
 @NgModule({
```
StojanovicMilos, line 59: I know it's a bit out of scope for this change, but why do we still use modules? When will we convert to standalone components? It's been the standard for a while

libs/entry-components/date-time-picker/time-picker.component.ts
```diff
@@ -12,6 +12,7 @@ export type meridiem = 'am' | 'pm';
 export class EntryTimePickerComponent<D> implements OnChanges {
   @HostBinding('class') class = 'entry-time-picker';
   readonly timeAdapter = inject(DateAdapter) as EntryDateTimeAdapter<D, unknown>;
+  private readonly changeDetectorRef = inject(ChangeDetectorRef);
```
StojanovicMilos, line 15: feels like this is going in the wrong direction - if we make the change detection on push and convert to signals, we wouldn't need this? @typhoon41 thoughts?

libs/entry-components/validation/entry-form-errors.component.ts
```diff
@@ -26,7 +27,27 @@ import { UntypedFormGroup } from '@angular/forms';
...
+
+  private readonly _changeDetectorRef = inject(ChangeDetectorRef);
+  private _statusSubscription: Subscription | undefined;
+
+  ngOnChanges(): void {
+    this._statusSubscription?.unsubscribe();
+
+    // `setServerSideValidationErrors` mutates the form in place, so the bound reference never
+    // changes and no event in this template ever marks this component dirty. Angular 22 made
+    // OnPush the default, so without this nothing re-renders the messages once they arrive.
```
StojanovicMilos, line 42: hint is in the comment, OnPush is the way to go, not this hacky approach.

libs/entry-form/tsconfig.lib.json
```diff
@@ -9,9 +9,9 @@
     "moduleResolution": "bundler",
     "types": [],
     "paths": {
-      "@enigmatry/entry-components/date-time-picker": ["dist/enigmatry/entry-components/date-time-picker"],
-      "@enigmatry/entry-components/common": ["dist/enigmatry/entry-components/common"],
-      "@enigmatry/entry-components/validation": ["dist/enigmatry/entry-components/validation"]
+      "@enigmatry/entry-components/date-time-picker": ["../../dist/enigmatry/entry-components/date-time-picker"],
```
StojanovicMilos, line 12: why is the path changed now? Is this something new in 22 or?

deploy-demo.yml
```diff
@@ -16,6 +16,13 @@ stages:
       fetchDepth: 50
       fetchTags: true
 
+    # Angular 22 requires Node ^22.22.3 || ^24.15.0 || >=26.0.0. Pinned so the build does not
+    # depend on whichever Node the hosted ubuntu-latest image happens to ship.
+    - task: NodeTool@0
```
StojanovicMilos: this whole task with comments above can be moved to separate yml to reduce duplication

libs/entry-components/permissions/README.md
```diff
@@ -55,6 +55,23 @@ providers: [
     }
 ```
 
+> **Angular 22 — check your `paramsInheritanceStrategy`**
+>
+> `entryPermissionGuard` reads `route.data['permissions']` and denies access when it is absent.
```
StojanovicMilos, line 60: entryPermissionGuard now fails open on child routes, and the mitigation is a README note. The guard is route.data['permissions'] ?? {} → return false when absent. Under 22's 'always' default a child with no permissions of its own inherits the parent's, so a route that was denied is now allowed to anyone who cleared the parent. Pinning 'emptyOnly' in the demo app is right and the permissions/README.md note is genuinely well written — but the library's authorization behavior now depends on a global router option the library doesn't control, and it changes silently for anyone who upgrades without reading a secondary entry-point readme. Worth deciding in the PR rather than deferring to the consumer:

read route.routeConfig?.data?.['permissions'] so the guard is strategy-independent — with the caveat that this isn't byte-identical to pre-22: under 'emptyOnly', a child with no data object at all did inherit, and would now be denied;
or at minimum a ngDevMode warning when a guarded route's permissions arrived from an ancestor.

libs/entry-components/README.md
```diff
@@ -31,6 +31,7 @@ These guides provides detailed steps for setting up and configuring theming with
 |19.x| = 19
 |20.x| = 20
 |21.x| = 21
+|22.x| = 22
```
StojanovicMilos, line 34: No migration notes anywhere. No CHANGELOG.md, and the READMEs only gained a |22.x| = 22 table row. Three things break consumers: the @angular/animations peer dep disappearing (they must drop provideAnimations()), the router default, and OnPush reaching every one of their own undeclared components. A top-level "Migrating to 22.x" section in libs/entry-components/README.md is the only channel they have.

tsconfig.json
```diff
@@ -2,25 +2,30 @@
...
-      "@enigmatry/entry-components/table": ["libs/entry-components/table/public-api.ts"],
-      "@enigmatry/entry-components/date-time-picker": ["libs/entry-components/date-time-picker/public-api.ts"],
-      "@enigmatry/entry-form": ["libs/entry-form/public-api.ts"],
-      "@enigmatry/entry-form/autocomplete": ["libs/entry-form/autocomplete/public-api.ts"],
-      "@enigmatry/entry-form/ckeditor": ["libs/entry-form/ckeditor/public-api.ts"],
-      "@enigmatry/entry-form/date-time-picker": ["libs/entry-form/date-time-picker/public-api.ts"]
+      // entry-codegen emits workspace-absolute imports (e.g. 'apps/demo-app/src/app/shared/shared.module')
+      // into apps/demo-app/**/generated/, which must not be hand-edited. This mapping replaces
+      // what `baseUrl` used to do for them.
+      "apps/*": ["./apps/*"],
```
StojanovicMilos, line 8: apps/* in the root tsconfig.json paths erodes the lib→app boundary. Every lib tsconfig extends the root and there's no no-restricted-imports or boundary rule in eslint.config.js, so libs/** can now import from apps/demo-app/** and compile clean. Only two files need the mapping, and one isn't generated: apps/demo-app/src/app/shared/services/file-load.service.ts:3 imports environment from 'apps/demo-app/src/environments/environment'. Make that relative and the mapping exists solely for products-generated.module.ts. Then add a lint rule to guard it — note that relocating it to tsconfig.app.json won't work cheaply, since paths isn't merged across extends and you'd have to re-declare all 15 @enigmatry/* entries.

package.json
```diff
@@ -57,18 +55,17 @@
     "@stylistic/eslint-plugin": "5.10.0",
     "@types/lodash-es": "^4.17.12",
     "@types/markdown-it": "^14.1.2",
-    "@types/node": "^24.1.0",
-    "angular-eslint": "21.3.1",
+    "@types/node": "^26.1.2",
```
StojanovicMilos: @types/node ^26.1.2 against a Node 24.x CI pin. Types describe APIs the runtime doesn't have. Pick one direction.

azure-pipelines.yml
```diff
@@ -21,6 +21,13 @@ stages:
       fetchDepth: 50
       fetchTags: true
 
+    # Angular 22 requires Node ^22.22.3 || ^24.15.0 || >=26.0.0. Pinned so the build does not
+    # depend on whichever Node the hosted ubuntu-latest image happens to ship.
+    - task: NodeTool@0
+      displayName: 'Install Node.js'
+      inputs:
+        versionSpec: '24.x'
```
StojanovicMilos: The Node pin exists only in CI. Your own skill note says it: "the repo declares no engines field, so this task is the only pin." Local devs get nothing, and neither do consumers of the published packages. Add engines to the root and to both published package.jsons. Also versionSpec: '24.x' is looser than the ^24.15.0 you documented as the requirement.

apps/demo-app/src/app/examples/validation/formly/complex-formly-form-validation/repeat-name.component.ts
StojanovicMilos: claude also saw this, although he's offering another direction which I don't like either @typhoon41 thoughts?

from claude: track $index in repeat-name.component.ts — that's a FieldArrayType, where children get added and removed. Removing a middle child is the one case index tracking gets wrong. track field (object identity, preserved by ngx-formly across remove()) is the safer key. field-set.component.ts is a static group, so $index there is harmless. Either way, a line on what forced the move off field.name (duplicate-key error?) would help — right now it reads like a default rather than a decision.

libs/entry-components/date-time-picker/time-picker.component.ts
```diff
@@ -79,5 +82,9 @@ export class EntryTimePickerComponent<D> implements OnChanges {
     if (this.meridiem === 'pm' && this.hours !== this.halfADay) {
       this.hours += this.halfADay;
     }
+
+    // Called by EntryDateTimePickerComponent from its calendarControl.valueChanges
```
StojanovicMilos: if we continue with this approach, here's more insight from claude:

time-picker comments are on the wrong one. The markForCheck() in update() is the non-obvious one — update() is called imperatively from the parent's (opened)="timePicker.update()", which marks the parent's view, not this child's — and it's uncommented. to24HourClock(), the more guessable case, got the comment.

libs/entry-components/date-time-picker/time-picker.component.ts
```diff
@@ -12,6 +12,7 @@ export type meridiem = 'am' | 'pm';
 export class EntryTimePickerComponent<D> implements OnChanges {
```
StojanovicMilos, line 12: date-time-picker.component.ts is asymmetric, and this PR is standing right next to it. Its formControl.statusChanges subscription calls markForCheck(); the valueChanges one doesn't, though it also updates calendarControl with emitEvent: false. A programmatic patchValue won't repaint. Pre-existing, cheap to fix here.

StojanovicMilos (changes requested):

apps/demo-app/src/app/examples/validation/formly/complex-formly-form-validation/repeat-name.component.ts
typhoon41: field.id?

libs/entry-components/date-time-picker/time-picker.component.ts
typhoon41: I agree.

libs/entry-form/tsconfig.lib.json
typhoon41: Deprecation in tsconfig - check baseUrl changes in it.

libs/eslint-config/package.json
```diff
@@ -21,9 +21,9 @@
     "eslint-plugin-prefer-arrow-functions": "3.9.1",
     "eslint-plugin-promise": "7.3.0",
     "eslint-plugin-unused-imports": "4.4.1",
-    "angular-eslint": "21.3.1",
-    "typescript-eslint": "8.59.1",
-    "eslint": "10.2.1"
+    "angular-eslint": "22.1.0",
+    "typescript-eslint": "8.66.0",
```
typhoon41: These changes must be done in a separate PR. Undo all eslint-related changes as those require special attention of what changed in all libraries we use.

deploy-demo.yml
typhoon41: You should be using UseNode@1 instead.

tsconfig.json
```diff
@@ -2,25 +2,30 @@
...
-      "@enigmatry/entry-components/button":["libs/entry-components/button/public-api.ts"],
-      "@enigmatry/entry-components/file-input":["libs/entry-components/file-input/public-api.ts"],
-      "@enigmatry/entry-components/permissions":["libs/entry-components/permissions/public-api.ts"],
-      "@enigmatry/entry-components/table": ["libs/entry-components/table/public-api.ts"],
-      "@enigmatry/entry-components/date-time-picker": ["libs/entry-components/date-time-picker/public-api.ts"],
-      "@enigmatry/entry-form": ["libs/entry-form/public-api.ts"],
-      "@enigmatry/entry-form/autocomplete": ["libs/entry-form/autocomplete/public-api.ts"],
-      "@enigmatry/entry-form/ckeditor": ["libs/entry-form/ckeditor/public-api.ts"],
-      "@enigmatry/entry-form/date-time-picker": ["libs/entry-form/date-time-picker/public-api.ts"]
+      // entry-codegen emits workspace-absolute imports (e.g. 'apps/demo-app/src/app/shared/shared.module')
```
typhoon41: Let's remove commented lines from this file. No real gain.

typhoon41 (changes requested):

apps/demo-app/src/app/examples/validation/formly/complex-formly-form-validation/field-set.component.ts
jovanasta: field.name wasn't merely non-unique — it was undefined on every item. name is an optional FormlyFieldConfig property whose only purpose is the HTML name attribute ("useful if you're posting the form to a server using techniques of yester-year", per Formly's own doc comment), and none of these field configs set it. So every item in each loop hashed to the same track key, and Angular logged NG0955 (duplicated keys) three times when the complex validation example rendered. Duplicate track keys also break @for diffing.

Pre-existing since the v21 upgrade — the @for migration wrote track field.name where the old *ngFor had no trackBy at all. Found while verifying v22 in the browser; the full reasoning is in commit e6fa1e9.

apps/demo-app/src/app/app-routing.module.ts
jovanasta: It will be done in one of the later pull requests

libs/entry-form/tsconfig.lib.json
jovanasta: It's TypeScript 6, not Angular 22. TS 6 removed baseUrl outright (TS5101), so it had to come out of the root tsconfig.json.

libs/eslint-config/package.json
jovanasta: done in part, one can't be undone. eslint 10.8.0 → 10.2.1 and typescript-eslint 8.66.0 → 8.59.1 are out (8.59.1's peer range is already typescript >=4.8.4 <6.1.0, so it never blocked TS 6). angular-eslint has to stay at 22.1.0: 21.3.1 peer-depends on @angular/cli >= 21.0.0 < 22.0.0 and pulls @angular-devkit/core/schematics in the same range, so pinning it back against CLI 22 either fails to resolve or drags a second devkit copy into the tree. angular-eslint majors track the Angular major one-to-one. If you want it fully isolated, the eslint PR has to land first against Angular 21 and this one rebase on top.

libs/entry-components/validation/entry-form-errors.component.ts
jovanasta: Will be done in later PR

libs/entry-components/date-time-picker/time-picker.component.ts
jovanasta: Will be done in later pr

libs/entry-components/file-input/entry-file-input.component.ts
```diff
@@ -40,6 +40,7 @@ const providers = [
 export class EntryFileInputComponent implements OnInit, OnDestroy, ControlValueAccessor, Validator {
     private readonly _ngZone: NgZone = inject(NgZone);
     private readonly _renderer: Renderer2 = inject(Renderer2);
+    private readonly _changeDetectorRef: ChangeDetectorRef = inject(ChangeDetectorRef);
```
typhoon41, line 43: Underscore in var name?

apps/demo-app/src/app/examples/validation/formly/complex-formly-form-validation/repeat-name.component.ts
typhoon41: Still unanswered — the `field.id?` suggestion above got no reply, and the code still uses `track $index` in both loops. This is the only review thread with consensus toward a change that hasn't landed: for this `FieldArrayType` (children added/removed), `$index` mis-tracks on middle removal, so `track field.id` (or `track field`) is the safer key. `field-set.component.ts` is a static group where `$index` is harmless, but it should probably stay consistent with whatever is decided here.

libs/entry-components/file-input/entry-file-input.component.ts
StojanovicMilos: it's all of them, not just changeDetectorRef - please remove the underscode in general

StojanovicMilos (approved):

typhoon41 (approved):

PR #291, opened 2026-08-14
BP-1582 - Migrate entry-components and entry-form to the signals API

jovanasta:
PR description

Migrates both published libraries and the demo app off decorator-based inputs, outputs and
queries, and off Angular lifecycle hooks, onto the signals API. Ships as part of **22.0.0** —
the first 22.x release — so the breaking changes below are in scope for the major.

## Breaking changes

Documented in full in `libs/entry-components/README.md` sections 5–8. Templates are
unaffected throughout; what changes is reading these members from TypeScript.

- **All lib inputs/outputs/queries are signals.** Imperative reads need `()`. A missed call
  is not a compile error in a boolean or template-literal position, so it fails quietly.
- `EntryDateTimePickerComponent.dateTimeChanged`: `Subject<D>` → `OutputEmitterRef<D>`.
  `.subscribe()` still works, `.next()` is gone.
- `EntryDialogComponent.confirm` / `cancel` → `confirmAction` / `cancelAction`.
  `[confirm]` / `[cancel]` bindings unchanged; the rename exists so the old call site is a
  *compile* error rather than a silent no-op.
- `EntryFileInputComponent`: `value` and `fileNames` are read-only signals; `disabled` is now
  the bound input only — the effective state (including the forms API) moved to
  `effectiveDisabled`.
- `NgControlAccessorDirective.control` is a getter with no setter.
- No lib component implements `ngOnInit` / `ngAfterViewInit` / `ngOnDestroy` any more — drop
  any `super.ngOnInit()` call.

## Verification

`npm run lint` clean, and `entry-components`, `entry-form` and `demo-app` all build. There is
no Angular test suite, so behaviour was verified by driving the demo app in Chrome: the
date-time picker round trip in both 24-hour and 12-hour/meridiem modes, reactive-form and
Formly validation, all five search filters plus filtering, the confirm dialog, spinner
overlay, auto-disable, throttled event plugins, and the regenerated API docs.

Two gaps, stated plainly:

- **The file-input Browse button was never clicked** — it opens a native file dialog. Its
  click-proxy listener moved from `ngOnInit` to `afterNextRender`, so that one path is
  compile-checked only. Worth a manual click.
- **The async-filter path is covered by construction, not in the browser.** The demo assigns
  its `searchFilters` in the constructor, which is precisely why a one-shot-input regression
  survived the first review round's browser pass.


apps/demo-app/src/app/shared/markdown-viewer/markdown-viewer.component.ts
```diff
@@ -1,53 +1,55 @@
...
+  private readonly renderer: Renderer2 = inject(Renderer2);
+  private readonly ngZone: NgZone = inject(NgZone);
+
+  /**
+   * Loads the documentation file whenever `fileUrl` is set. Replaces the ngOnInit subscription; the
+   * resource reloads if a different url is bound, which the hook did not.
+   */
+  private readonly loadedFile = resource({
+    params: () => this.fileUrl(),
+    loader: ({ params: fileUrl }) => firstValueFrom(this.fileLoad.loadDocumentationFile(fileUrl))
```
StojanovicMilos: why not just have this.fileUrl in loader? Also, why not async loader?

libs/entry-components/common/directives/ng-control-accessor.directive.ts
```diff
@@ -1,43 +1,62 @@
...
+     * Control the host is bound to through the forms API, or a standalone control when the host is
+     * not bound to one.
+     *
+     * @remarks Resolved per read rather than memoised on the first one. `FormControlName._control`
+     * and `FormControlDirective.form` are only populated once the owning form directive has run its
+     * `ngOnChanges`, so an early read genuinely has nothing to return - memoising it would either
+     * cache `undefined` behind a non-nullable type or lock in the standalone fallback for good. Only
+     * the fallback is cached, so its identity is stable for as long as it is in use.
+     */
+    get control(): UntypedFormControl {
```
StojanovicMilos: I thought we aren't going to add UntypedFormX anymore? Or getters?

libs/entry-components/common/directives/ng-control-accessor.directive.ts
```diff
@@ -1,43 +1,62 @@
...
-    }
-
-    ngOnDestroy(): void {
-        this.destroy$.next();
-        this.destroy$.complete();
-    }
-}
\ No newline at end of file
+        const ngModel = this.ngControl;
+        ngModel.control.valueChanges
```
StojanovicMilos: signals?

libs/entry-components/date-time-picker/date-time-picker.component.ts
```diff
@@ -17,36 +17,33 @@ import { EntryTimePickerComponent } from './time-picker.component';
     ],
     hostDirectives: [NoopControlValueAccessorDirective, NgControlAccessorDirective],
     changeDetection: ChangeDetectionStrategy.OnPush,
-    standalone: false
+    standalone: false,
```
StojanovicMilos: do we have in the future todos to switch to standalone?

libs/entry-components/date-time-picker/time-picker.component.ts
```diff
@@ -7,87 +7,99 @@ export type meridiem = 'am' | 'pm';
...
-  @Input() showSeconds: boolean;
-  @Input() is12HourClock: boolean;
-  @Input() defaultTime: D | undefined;
+  readonly date = input<D | undefined>(undefined);
+  readonly showSeconds = input(false);
+  readonly is12HourClock = input(false);
+  readonly defaultTime = input<D | undefined>(undefined);
+
+  /** Bumped by `update()`, whose job is to re-read `today()` - a value signals cannot track. */
+  private readonly refreshCount = signal(0);
```
StojanovicMilos: Generally not a best practice, but I see why you did it. Since you already have to do it, let's at least protect from theoretical overflow - convert it to boolean and just flip it or when you increment it, have moduo at some number so it can never reach number's max value.

libs/entry-components/dialog/dialogs/entry-dialog.component.ts
```diff
@@ -16,37 +16,53 @@ import { ENTRY_DIALOG_CONFIG, EntryDialogConfig } from '../entry-dialog-config.m
...
-    @Input() disableConfirm: boolean;
+    readonly disableConfirm = input(false);
     /** Provide custom buttons template */
-    @Input() buttonsTemplate: TemplateRef<any> | null | undefined;
+    readonly buttonsTemplate = input<TemplateRef<any> | null | undefined>(undefined);
 
-    @Input() confirm: () => Observable<unknown> = () => of(true);
-    @Input() cancel = () => this.close(false);
+    /**
+     * Callback invoked when the dialog is confirmed.
```
StojanovicMilos: too many comments such as this one which are too verbose. AI in general overdoes it with comments that get stale fast and/or nobody reads.

libs/entry-components/dialog/dialogs/entry-dialog.component.ts
```diff
@@ -16,37 +16,53 @@ import { ENTRY_DIALOG_CONFIG, EntryDialogConfig } from '../entry-dialog-config.m
...
+     *
+     * @remarks Bound as `[confirm]`, but named `confirmAction` on the class. `confirm` used to be a
+     * directly callable member; as a signal input `this.confirm()` would return the callback rather
+     * than invoke it, which type-checks fine and fails silently for anyone extending this component.
+     * The alias keeps the template binding while turning that into a compile error - which is why
+     * no-input-rename is waived here rather than the alias dropped.
+     */
+    // eslint-disable-next-line @angular-eslint/no-input-rename
+    readonly confirmAction = input<() => Observable<unknown>>(() => of(true), { alias: 'confirm' });
```
StojanovicMilos: at some point (not in this PR/ticket) we should probably brainstorm how to convert this to generic so that we don't have unknown that can be overridden with whatever.

StojanovicMilos: alias? Why?

libs/entry-form/autocomplete/check-autocomplete-input.directive.ts
```diff
@@ -1,58 +1,57 @@
...
+    });
   }
 
   private checkControlValue(): void {
     const controlValue = this.control?.value;
     if (!controlValue) {
       return;
     }
-    if (findOptionByValue(this.options, controlValue)) {
+    if (findOptionByValue(this.options() ?? [], controlValue)) {
```
StojanovicMilos: options has default value and is not nullable, so it should never reach ?? []?

StojanovicMilos (changes requested):

apps/demo-app/src/app/examples/search-filter/search-filter/search-filter-example.component.ts
```diff
@@ -10,19 +10,20 @@ import {
...
 @Component({
     selector: 'app-search-filter-example',
     templateUrl: './search-filter-example.component.html',
-    styleUrls: ['./search-filter-example.component.scss'],
+    styleUrl: './search-filter-example.component.scss',
     standalone: false
 })
 export class SearchFilterExampleComponent {
-  @ViewChild(EntrySearchFilterComponent, { static: true }) entrySearchFilterComponent: EntrySearchFilterComponent;
+  // Not `.required`: the constructor kicks off a fetch whose error handler can run before the view
```
typhoon41: This comment is really off...and ofc redundant.

apps/demo-app/src/app/shared/example-viewer/code-view/code-view.component.ts
```diff
@@ -1,39 +1,50 @@
...
+    try {
+      return hljs.highlight(code, { language }).value;
+    } catch {
+      return code;
+    }
+  };
 
-    this.highlightedCode = this._domSanitizer.bypassSecurityTrustHtml(sanitizedHtml || '');
-  }
+  copy = () => {
```
typhoon41: readonly?

apps/demo-app/src/app/shared/example-viewer/example-viewer.component.ts
```diff
@@ -33,30 +33,27 @@ export class ExampleViewerComponent implements OnDestroy {
...
+      typescript: this.showTs() ? this.loadFile(component, 'ts') : of(null),
+      html: this.showHtml() ? this.loadFile(component, 'html') : of(null),
+      styles: this.showScss() ? this.loadFile(component, 'scss') : of(null),
+      docs: this.showDocs() ? this.loadFile(component, 'md') : of(null)
     })
       .pipe(
-        takeUntil(this._destroy$)
+        takeUntilDestroyed(this.destroyRef)
       )
       .subscribe(documents => {
```
typhoon41: We will try to eliminate subscriptions in some of the consecutive PRs?

apps/demo-app/src/app/shared/landing/landing.component.ts
```diff
@@ -7,23 +7,23 @@ import { IComponentDefinition, COMPONENT_DEFINITIONS } from '../../features/comp
...
 })
 export class LandingComponent {
   menuItems = COMPONENT_DEFINITIONS;
 
-  private readonly _router: Router = inject(Router);
-  private _activatedRoute: ActivatedRoute = inject(ActivatedRoute);
-  private _clipboard: Clipboard = inject(Clipboard);
-  private _snackBar: MatSnackBar = inject(MatSnackBar);
+  private readonly router: Router = inject(Router);
+  private activatedRoute: ActivatedRoute = inject(ActivatedRoute);
```
typhoon41: why these aren't readonly?

apps/demo-app/src/app/shared/markdown-viewer/markdown-viewer.component.ts
```diff
@@ -1,53 +1,55 @@
...
+  });
+
+  /** Inline content wins over a loaded file, and a failed load falls back to a notice. */
+  readonly markdownContentHtml = computed<SafeHtml>(() => {
```
typhoon41: We can also instruct Claude to generate protected readonly for all public methods/signals/properties that are used only in markup (but not in other components/services).

apps/demo-app/src/app/shared/markdown-viewer/markdown-viewer.component.ts
```diff
@@ -1,53 +1,55 @@
...
-      .loadDocumentationFile(this.fileUrl!)
-      .pipe(
-        map(response => this.convertMarkdownToHtml(response))
-      )
-      .subscribe({
-        next: response => this.markdownContentHtml.set(response),
-        error: _ => this.markdownContentHtml.set(`### No API documentation found :'(`)
-      });
+  constructor() {
+    // Replaces ngOnInit for the delegated click handling - the host element has to be rendered first.
```
typhoon41: Remove this redundant comment.

libs/entry-components/common/directives/ng-control-accessor.directive.ts
```diff
@@ -1,43 +1,62 @@
...
 import { FormControlDirective, FormControlName, NgControl, NgModel, UntypedFormControl } from '@angular/forms';
-import { Subject, takeUntil } from 'rxjs';
 
 @Directive({
     standalone: true
 })
-export class NgControlAccessorDirective implements OnDestroy, OnInit {
-    control: UntypedFormControl;
-
+export class NgControlAccessorDirective {
```
typhoon41: Is there still a justification for existence of this class in signal era? If it's temporary, in which exact step is time for it to go?

libs/entry-components/common/directives/scroll-to-invalid-control.directive.ts
```diff
@@ -1,47 +1,46 @@
...
+/**
+ * Scroll to first invalid control when form is submitted.
+ * Directive is applied to 'form[formGroup],form[ngForm]' (reactive or template driven forms)
+ */
+@Directive({
+  standalone: true,
+  // eslint-disable-next-line @angular-eslint/directive-selector
+  selector: 'form[formGroup],form[ngForm]'
+})
+export class ScrollToInvalidControlDirective {
```
typhoon41: Reconsider existence of this class with Claude. How can it be removed/improved?

libs/entry-components/dialog/dialogs/entry-dialog.component.ts
typhoon41: AI reads it :D

libs/entry-components/search-filter/autocomplete/autocomplete-search-filter.component.ts
```diff
@@ -12,50 +12,57 @@ import { AutocompleteSearchFilter } from './autocomplete-search-filter.model';
...
+    // read after the input was set. Going through the input signal defers that instead.
+    toObservable(this.searchFilter)
       .pipe(
-        takeUntil(this.destroy$),
-        tap(value => this.clearFilterIfLabelMismatch(value)),
-        filter(value => !!value && value.length >= this.searchFilter.minimumCharacters),
-        debounceTime(this.searchFilter.debounceTime)
+        switchMap(searchFilter => this.searchField.valueChanges.pipe(
+          tap(value => this.clearFilterIfLabelMismatch(value)),
+          filter(value => !!value && value.length >= searchFilter.minimumCharacters),
```
typhoon41: Effects, signals?

typhoon41:
Two P2 findings from a Codex review of this branch:

**[P2] Attach form subscriptions before parent view hooks run** — `libs/entry-components/date-time-picker/date-time-picker.component.ts:76-80`
When a parent disables the bound control in `ngAfterViewInit`, this `afterNextRender` callback has not subscribed yet and `statusChanges` does not replay, leaving `calendarControl` enabled and its date-picker toggle interactive. The previous `ngOnInit` subscription observed such changes; subscribe earlier or seed the mirror from the control's current status when attaching.

**[P2] Use a collision-free filter-set cache key** — `libs/entry-components/search-filter/entry-search-filter.component.ts:63`
When a dynamic filter set changes from one key containing `|` (for example `['name|email']`) to multiple keys such as `['name', 'email']`, both arrays produce the same cache key, so the form is not rebuilt and the old controls remain rendered. Since filter keys are unrestricted strings, compare the key arrays structurally or use collision-free serialization.

libs/entry-components/search-filter/entry-search-filter.component.ts
```diff
@@ -1,75 +1,109 @@
...
+   */
+  readonly searchFilterChange = output<SearchFilterParams>();
+
+  private built: { keys: string; filters: SearchFilterBase<any>[]; form: UntypedFormGroup } | undefined;
+
+  /**
+   * The filters the form was actually built from. The template iterates this rather than the input
+   * so the rendered controls and the form can never disagree.
+   */
+  get renderedSearchFilters(): SearchFilterBase<any>[] {
```
typhoon41: Getters?

libs/entry-components/search-filter/entry-search-filter.component.ts
```diff
@@ -1,75 +1,109 @@
...
+  readonly config: EntrySearchFilterConfig = inject(ENTRY_SEARCH_FILTER_CONFIG);
+  private destroyRef = inject(DestroyRef);
+
+  readonly onSubmit = (): void => {
+    const formValue = this.searchFilterForm.value;
+    this.searchFilterChange.emit(formValue);
+  };
+
+  toFormGroup = (searchFilters: SearchFilterBase<any>[]) => {
+    const group: any = {};
```
typhoon41: no anys should be in our codebase once we touch it!

typhoon41 (changes requested):

libs/entry-components/date-time-picker/date-time-picker.component.ts
```diff
@@ -59,77 +56,95 @@ export class EntryDateTimePickerComponent<D> implements OnInit, OnDestroy {
...
+    // which is what the ngOnInit call to setDisabled used to cover.
+    effect(() => this.setDisabled(this.disabled()));
+
+    // Replaces ngOnInit. The forms API has populated `formControl` by the time the host has
+    // rendered, and calendarControl drives a hidden input, so nothing is visibly late.
+    afterNextRender(() => {
+      this.calendarControl.setValue(this.formControl.value, { emitEvent: false });
+      this.mirrorDisabledState();
+      this.mirrorValueChanges();
+      this.applyCalendarSelection();
```
typhoon41: **[P2] Attach form subscriptions before parent view hooks run**

When a parent disables the bound control in `ngAfterViewInit`, this `afterNextRender` callback has not subscribed yet and `statusChanges` does not replay, leaving `calendarControl` enabled and its date-picker toggle interactive. The previous `ngOnInit` subscription observed such changes; subscribe earlier or seed the mirror from the control's current status when attaching.

libs/entry-components/search-filter/entry-search-filter.component.ts
```diff
@@ -1,75 +1,109 @@
...
+   * run. Building strictly once instead loses filters that arrive after the first render, which is
+   * the normal shape for a set assembled in an HTTP callback - and loses them silently, because the
+   * template renders this snapshot.
+   *
+   * Same keys, different instances: the first instances win, so a re-bound filter's `label` or
+   * `placeholder` change is not picked up. Changing the key set is the supported way to swap filters.
+   */
+  private readonly build = (): { keys: string; filters: SearchFilterBase<any>[]; form: UntypedFormGroup } => {
+    const filters = this.searchFilters();
+    const keys = filters.map(searchFilter => searchFilter.key).join('|');
```
typhoon41: **[P2] Use a collision-free filter-set cache key**

When a dynamic filter set changes from one key containing `|` (for example `['name|email']`) to multiple keys such as `['name', 'email']`, both arrays produce the same cache key, so the form is not rebuilt and the old controls remain rendered. Since filter keys are unrestricted strings, compare the key arrays structurally or use collision-free serialization.

typhoon41 (commented):
Posting two P2 findings from a Codex review of this branch.

apps/demo-app/src/app/shared/markdown-viewer/markdown-viewer.component.ts
jovanasta: Async: done.

The `fileUrl` part has to stay in `params` though — a signal read inside a `resource` loader is untracked, so the resource would fetch once and never reload when a different url is bound. `params` is also what keeps it idle for the `[markdownContent]` usage: it returns `undefined` there, and a resource with undefined params never calls the loader at all.

One oddity worth flagging: the loader is `async` without an `await`, because our own `@typescript-eslint/return-await` rejects `return await` in that position.

libs/entry-components/common/directives/ng-control-accessor.directive.ts
jovanasta: Fair — and gone. There is no `Untyped*` left anywhere in `libs/`.

It wasn't a rename, because `UntypedFormGroup` *is* `FormGroup<any>` — swapping the name would have bought nothing:

- the search-filter form is now a `FormRecord` (dynamic keys, controls typed `AbstractControl`)
- `setServerSideValidationErrors` and `entry-form-errors` take `AbstractControl` — both **widenings**, so you can now pass a fully typed `FormGroup` where you previously had to widen
- `NgControlAccessorDirective.control` is `AbstractControl`; the date-time picker casts to `FormControl<D>` where the type is actually known

On the getter: it is load-bearing. The owning form directive only populates its control in `ngOnChanges`, so a value captured at construction is either `undefined` or the standalone fallback, permanently. Only the fallback is cached.

The one remaining `UntypedFormControl` in the picture is Formly's own `FieldType.formControl`, which is not ours to change. Documented in README §7.

libs/entry-components/common/directives/ng-control-accessor.directive.ts
jovanasta: Signals aren't the answer here, but neither is keeping this — my first take was too shallow, so here is the real picture.

`EntryDateTimePickerComponent` does not implement `ControlValueAccessor` at all. It provides a **no-op** CVA (`NoopControlValueAccessorDirective`) purely to satisfy Angular's requirement that a `[formControl]`-bound element have one, then uses `NgControlAccessorDirective` to reach around it and grab the real control, and writes to it directly.

`keepNgModelInSync` exists *only* to patch up the `ngModel` case that the no-op CVA breaks. So the right fix is not to convert this subscription — it's to delete it, by implementing `ControlValueAccessor` properly on the picker. See my reply on the thread above about this class's lifetime.

libs/entry-components/date-time-picker/date-time-picker.component.ts
jovanasta: Not in this ticket. CLAUDE.md now records the actual state instead of the assumption we had: 69 of 91 components/directives/pipes still declare `standalone: false`, and Formly was never the blocker — `entry-components` has no `@ngx-formly` imports at all. The only piece needing design work is `FormlyAutocompleteModule`, which registers its field type through `FormlyModule.forChild()`.

@jovanasta is confirming whether a follow-up ticket exists or needs raising.

libs/entry-components/dialog/dialogs/entry-dialog.component.ts
jovanasta: Agreed, and agreed it's not this PR. Leaving this open as the placeholder for that discussion.

jovanasta: Because `confirm` used to be directly callable.

As a plain signal input, `this.confirm()` returns the callback instead of invoking it — that type-checks fine and fails silently for any subclass. Renaming the class member to `confirmAction` turns it into a compile error, and the alias keeps `[confirm]` working in every existing template, so consumers see no breakage.

It's the same trap README §8 documents for `EntryFileInputComponent.disabled`. If you'd rather drop the alias and rename the binding too, that's a bigger break for consumers — happy to go that way if you prefer it.

apps/demo-app/src/app/shared/example-viewer/code-view/code-view.component.ts
jovanasta: Done — `protected readonly`, since it's only reached from this component's own template.

apps/demo-app/src/app/shared/example-viewer/example-viewer.component.ts
jovanasta: Yes. What's left in the demo app is the `forkJoin(...).subscribe()` here and the `ValidationService` error-path subscriptions — both one-shot, so both should be `firstValueFrom` + `await` per our own convention. Left out of this PR to keep it to the signals migration.

apps/demo-app/src/app/shared/landing/landing.component.ts
jovanasta: Fixed — these are `readonly` now, and the template-only members are `protected readonly`.

apps/demo-app/src/app/shared/markdown-viewer/markdown-viewer.component.ts
jovanasta: Good idea — added to CLAUDE.md as a "Member visibility" convention, so it applies going forward rather than just here.

One caveat recorded with it: in `libs/` this only applies to genuinely internal members. These are published packages, so narrowing an exported class's member from `public` to `protected` removes it from the published API and breaks consumers — that's an API decision, not a style one. Applied throughout the demo app.

apps/demo-app/src/app/shared/markdown-viewer/markdown-viewer.component.ts
jovanasta: Removed. Trimmed the over-long comment blocks across the branch too, per Miloš's note.

libs/entry-components/common/directives/ng-control-accessor.directive.ts
jovanasta: Good question, and the honest answer is that it shouldn't exist long-term — but the reason is more specific than "signal era".

`EntryDateTimePickerComponent` does not implement `ControlValueAccessor`. It provides a **no-op** CVA just to satisfy Angular's requirement that a `[formControl]`-bound element have one, then uses this directive to reach around it and grab the real control, writing to it directly. `keepNgModelInSync` then exists only to patch up the `ngModel` case the no-op CVA breaks.

**The exact step is: implement `ControlValueAccessor` properly on the date-time picker.** That deletes all three pieces — the no-op CVA, this directive, and the ngModel sync. It's a real refactor rather than a syntax change, so it wants its own ticket.

Signal forms would also remove it (`@angular/forms/signals` is public API in 22.0, with a `compat` bridge), but they are not the prerequisite — this is fixable today.

libs/entry-components/common/directives/scroll-to-invalid-control.directive.ts
jovanasta: I'd keep it, but you're right that it has a gap — and I don't have an "it's fine as-is" answer.

It scrolls to the first invalid control but never focuses it, so screen-reader and keyboard users get no indication anything happened (WCAG 2.4.3 / 3.3.1). One line inside `scrollToInvalidControl`:

```ts
firstInvalidControl.scrollIntoView({ behavior: 'smooth', block: 'center' });
firstInvalidControl.focus({ preventScroll: true });
```

Not applied yet — it changes runtime behaviour and this PR has had three review rounds already. Say the word and it goes in, otherwise it's a clean follow-up.

libs/entry-components/search-filter/autocomplete/autocomplete-search-filter.component.ts
jovanasta: The result already lands in a signal — that's what marks the OnPush view. The pipeline itself stays RxJS because `debounceTime` and switch-cancellation of an in-flight search have no signal equivalent, and `resource` doesn't cover it either (no debounce). Added a two-line note saying exactly that, so the next reader doesn't have to ask.

libs/entry-components/search-filter/entry-search-filter.component.ts
jovanasta: Fixed — `renderedSearchFilters` and `searchFilterForm` are `computed` signals now.

That also fixed a leak I hadn't spotted: the getters ran `toFormGroup` once per binding per change-detection pass, and the `formatValue` subscriptions from previous builds were never torn down. Both gone.

Note `searchFilterForm` is a signal now, so imperative reads need `searchFilterForm()` — added to the README migration table, since that's how server-side validation errors get onto the filter form.

libs/entry-components/search-filter/entry-search-filter.component.ts
jovanasta: Agreed in principle, but this one can't be a straight swap and I'd rather flag it than quietly pick.

I tried `SearchFilterBase<any>[]` → `SearchFilterBase<unknown>[]`. The library compiled — but only because nothing *inside* it assigns a concrete subclass. The demo app then failed on every filter:

```
TS2322: Type 'TextSearchFilter' is not assignable to type 'SearchFilterBase<unknown>'.
  Types of property 'formatValue' are incompatible.
    Type '(value: string) => string' is not assignable to type '(value: unknown) => unknown'.
```

`formatValue` is a function-valued **property**, so under `strictFunctionTypes` its parameter is contravariant — `SearchFilterBase<string>` is not assignable to `SearchFilterBase<unknown>`. It's the only contravariant member, so it's the whole blocker.

Two routes:

**A** — `export type AnySearchFilter = SearchFilterBase<any>` with the variance reason documented, used in the four places. Not breaking; `any` reduced to one named, explained occurrence.

**B** — retype `formatValue` as `(value: unknown) => unknown`. Actually removes the `any`, but it's a **breaking change**: consumers writing `formatValue: (v: string) => v.trim()` must narrow inside the callback. Our own `maskDecimalScore(value: string)` is exactly that shape.

B is what you asked for literally; A is what doesn't break consumers mid-release. Which do you want? Leaving this open.

libs/entry-components/date-time-picker/date-time-picker.component.ts
jovanasta: Real one — fixed. `mirrorDisabledState` now seeds from the current status:

```ts
.pipe(startWith(this.formControl.status), takeUntilDestroyed(this.destroyRef))
```

Verified in the browser on the "with disabling" example: disabling via the forms API and via `[disabled]`+`ngModel` both propagate to `calendarControl` and the date-picker toggle, and both reverse.

libs/entry-components/search-filter/entry-search-filter.component.ts
jovanasta: Real one — fixed. The joined-string key is gone; the cache now compares the key arrays element-wise, as the `equal` of a `computed`:

```ts
private readonly stableFilters = computed(() => this.searchFilters(), {
  equal: (a, b) => a.length === b.length && a.every((filter, i) => filter.key === b[i].key)
});
```

No serialisation, so no collision is possible regardless of what the keys contain.

libs/entry-components/common/directives/ng-control-accessor.directive.ts
typhoon41: Not sure what to think of it; if it's really tens of new files changed, I'd say do it in a different ticket, otherwise let's do this too.

This is what Codex said on this: "Claim doesn't match the code. The host-directive relationship is visible in EntryDateTimePickerComponent, but neither NgControlAccessorDirective nor the date-time picker documents the claimed no-op-CVA rationale or long-term removal path."

typhoon41 (changes requested):

libs/entry-components/common/directives/ng-control-accessor.directive.ts
typhoon41: From the Codex re-review: the documentation this thread refers to didn't land — neither `NgControlAccessorDirective` nor `EntryDateTimePickerComponent` documents the no-op-CVA rationale or the long-term removal path. The host-directive relationship is visible in the code, but the "why" from this thread only exists here in the PR conversation. Please record it in the code.

libs/entry-components/common/directives/scroll-to-invalid-control.directive.ts
typhoon41: From the Codex re-review: the focus call described here was not added — `scroll-to-invalid-control.directive.ts:39` still only calls `scrollIntoView`, so keyboard and screen-reader focus stays on the submit control. Please add the `focus()` call this thread agreed on.

libs/entry-components/search-filter/entry-search-filter.component.ts
typhoon41: From the Codex re-review: confirming this is still open — `SearchFilterBase<any>[]` remains at `entry-search-filter.component.ts:27` and in the demo. The variance analysis above matches what landed, so if we deliberately keep `any` here, add a brief comment at the declaration explaining why `unknown` doesn't work, so the next sweep doesn't re-litigate it.

libs/entry-components/search-filter/entry-search-filter.component.ts
```diff
@@ -1,75 +1,106 @@
...
+  /** Torn down and replaced on every rebuild, so old controls do not keep formatting subscriptions alive. */
+  private formatSubscriptions = new Subscription();
+
+  /**
+   * The input, but only changing identity when the set of filter keys does. Rebuilding on array
+   * identity would discard whatever the user typed whenever a caller binds `[searchFilters]="getFilters()"`,
+   * since `toFormGroup` mints new controls; building strictly once would silently drop filters that
+   * arrive after the first render.
+   */
+  private readonly stableFilters = computed(() => this.searchFilters(), {
```
typhoon41: **[P2]** From the Codex re-review: in-place changes to the filter array no longer rebuild the form. This caches through an input signal, and Angular won't invalidate it when a caller mutates the existing array (e.g. `this.filters.push(new TextSearchFilter(...))`). The previous getter inspected the keys on every read; now a pushed filter is neither rendered nor added to the form. Either preserve mutation support or explicitly document that the input requires immutable replacement (`this.filters = [...this.filters, newFilter]`).

libs/entry-components/search-filter/entry-search-filter.component.ts
```diff
@@ -1,75 +1,106 @@
...
+    this.formatSubscriptions = new Subscription();
+
+    return { filters, form: this.toFormGroup(filters) };
+  });
+
+  /**
+   * The filters the form was actually built from. The template iterates this rather than the input
+   * so the rendered controls and the form can never disagree.
+   */
+  readonly renderedSearchFilters: Signal<SearchFilterBase<any>[]> = computed(() => this.built().filters);
```
typhoon41: **[P3]** From the Codex re-review: this is an undocumented public API break — `renderedSearchFilters` changed from an array property to `Signal<SearchFilterBase<any>[]>`, so consumers doing `component.renderedSearchFilters.map(...)` no longer compile and must call `renderedSearchFilters()`. The 22.0.0 migration table documents `searchFilterForm` but omits this member — please add it.

libs/entry-components/permissions/permission.directive.ts
```diff
@@ -7,22 +7,58 @@ import { EntryPermissionService } from './permission.service';
...
+    effect(() => this.toggleVisibility(!this.isHidden()));
   }
 
   private readonly toggleVisibility = (show: boolean): void => {
     const element = this.elementRef.nativeElement;
 
-    if(!show) {
+    // SVG hosts must still be hidden, so `HTMLElement` alone is too narrow. Only the comment node
+    // produced by the unsupported `*entryPermissionsOnly` form falls through here.
+    if (!(element instanceof HTMLElement) && !(element instanceof SVGElement)) {
```
typhoon41: **[P2]** From the Codex re-review: this can crash server-side rendering. `HTMLElement` and `SVGElement` are browser globals; in a Node SSR environment they're undefined, so rendering any permission-gated element throws `ReferenceError` before `Renderer2` gets a chance to act. Use a platform-neutral check (e.g. `element.nodeType === 1`, or guard with `typeof HTMLElement !== 'undefined'`) instead of the bare DOM constructors.

libs/entry-components/permissions/permission.directive.ts
```diff
@@ -7,22 +7,58 @@ import { EntryPermissionService } from './permission.service';
...
+
+  /** Hides the host element when the user holds these permissions. */
+  readonly except = input<T[] | undefined>(undefined, { alias: 'entryPermissionsExcept' });
+
+  /**
+   * `only` wins when both aliases are bound, which the selector allows but no caller does.
+   * Denies by default: an unresolved binding arrives as `undefined`, and for a permission gate
+   * "nothing to check against" has to read as hidden.
+   */
+  private readonly isHidden = computed(() => {
```
typhoon41: **[P2]** From the Codex re-review: binding both inputs changed behavior — `entryPermissionsOnly` now always wins, whereas previously both input setters toggled visibility, so the last applied binding determined the result. For a user holding both `read` and `suspended`, an element bound with only=`read` + except=`suspended` can now be shown because the exclusion is ignored. Either reject simultaneous bindings (report via ErrorHandler like the unsupported-host case) or define combined semantics that evaluate both conditions.

libs/entry-components/search-filter/entry-search-filter.component.ts
typhoon41: I want option B, of course. We're introducing bunch of breaking changes, blood pressure 120mmHg over 80mmHg.

StojanovicMilos (commented):
## Modern-Angular follow-ups from the signals migration

Checked the branch against the shipped v22 typings (`@angular/core`, `@angular/forms`, `@angular/material` 22.1.3) rather than the docs prose, which is stale in places. Items 4, 5, 8, 11 and 12 are being handled in this PR, along with two of the inline P2s; the rest is separate work. The inline P1 on the search-filter rebuild is deferred into the signal-forms ticket rather than fixed twice — see item 6.

### 1. No Angular test suite — highest priority - Status: Separate ticket, agreed to wait until a functional v22 has shipped.
**Evidence:** this migration touched 75 files and every published input/output/query in both libraries, and was verified by driving the demo app in Chrome. The PR body itself names two paths that were never exercised (the file-input Browse click, the async-filter path). A breaking major of a published library with zero automated coverage is the largest risk on this branch.
**Ticket:** *Add a test suite to entry-components and entry-form*

### 2. No CI on pull requests - Status: Withdrawn as written; see the correction below.
**Evidence:** zero check runs on the head commit. "`npm run lint` clean, three builds green" is author-attested only and cannot be re-verified by a reviewer or after a rebase.
**Ticket:** *Add a PR pipeline running lint + build for both libs and the demo app* — pairs with #1.

Correction to item above — I was wrong that there's no CI. There is: ADO project Enigmatry - Angular Building Blocks, definition 31 entry-angular-building-blocks, which runs npm ci, lint, both library builds and the SCSS tests before publishing to npm.

What's actually true is narrower. azure-pipelines.yml has pr: none, and every recent run of definition 31 is on refs/heads/master — so a pull request gets no check run, and a break surfaces after merge as a failed publish rather than as a blocked merge. Whether that's worth changing is a separate conversation; striking the item as written.

There is a sharper version of this that I missed first time round, and it is worse than a failed publish: every master build publishes with no dist-tag argument, so a MinVer preview takes latest. On 23 June, 20.3.4-preview.1 and 20.3.6-preview.1 published after 21.2.4 and latest moved backwards to a 20.x preview. A consumer pinned to ^21 is safe by semver; a fresh install with no pin is not. That is its own ticket — publish previews under next and reserve latest for tagged releases — and it wants doing before this PR merges, not after.

### 3. Standalone migration - Status: Separate ticket, after the breaking set. entry-components can go first — it has no @ngx-formly imports; only entry-form's FormlyModule.forChild() registration needs the Formly removal.
**Evidence:** 69 declarables still carry `standalone: false`. This PR's own CLAUDE.md update establishes that Formly is **not** the blocker it was assumed to be — `entry-components` has no `@ngx-formly` imports at all, and a standalone component can be imported into an NgModule. The one piece needing design work is `FormlyAutocompleteModule` and its `FormlyModule.forChild()` registration.
**Ticket:** *Migrate entry-components and entry-form to standalone* (already promised in the date-time-picker thread — please link the number here once created).

### 4. prefer-standalone is disabled for a reason that is no longer true — fix in this PR - Status: Done in this PR — 60154b2 and 06cb04c.

Evidence: the same stale comment sat in two places — eslint.config.js:12 for the libraries, and apps/demo-app/eslint.config.js:9 for the demo app, which is linted from its own directory with its own config. Both read // TODO: Remove when we get rid of Formly, which the CLAUDE.md text added on this branch directly contradicts. Formly is not the blocker in either place: a standalone component can import a Formly module.
Disposition: both comments now point at the standalone work; re-enable the rule as part of #3.

### 5. Lint rules that would catch the silent failure the migration guide warns about - Status: no-uncalled-signals and restrict-template-expressions are enabled in this PR — bca60fd and b7666be, zero violations each. prefer-signals and prefer-signal-model go to a separate ticket.

Evidence: README §5 says a missed () "is not a compile error in a boolean or template-literal position, so it fails quietly". Both positions are now build errors, and neither cost a new dependency:

@angular-eslint/no-uncalled-signals — the boolean half: if (signal), comparisons, negation, ternaries, both operands of &&. It does not catch ${signal}.
@typescript-eslint/restrict-template-expressions — the interpolation half, reported as an invalid template literal type.
@angular-eslint/prefer-signals and prefer-signal-model — still to do; they flag remaining @Input/@Output/@ViewChild and two-way pairs, and will surface their own pile of violations.

Both were confirmed by probe rather than assumed, since a rule with zero violations looks identical to an inert one — and that mattered here, because the demo app lints from its own directory and a root-only run never sees it.
Ticket: Enable prefer-signals, prefer-signal-model and the remaining type-aware lint backstops

### 6. Signal forms - Status: Separate ticket, in v22, piloted on EntrySearchFilterComponent — and now first in the breaking set, because the inline P1 on the search filter is deferred into it.
**Evidence:** signal forms are **stable** in `@angular/forms@22.1.3` — `form()`, `FieldTree`, the `[formField]` directive and all validators ship with no `@experimental` marker (the tutorial page on angular.dev still says otherwise; the typings are the authority). There is also a `@angular/forms/signals/compat` entry point for interop with reactive forms.

This matters more than a generic modernization item, because the search filter is where reactive forms are actively hurting: it keeps form state in two places and has been rewritten four times across four review rounds — getters that rebuilt on every read and leaked subscriptions, then a string cache key with a collision, then key-equality that silently dropped a rebound array, and now control state discarded on every change-detection pass. Every one of them assumes the form is derived from the input and then argues about when to recompute, and there is no cache key that is simultaneously collision-free, mutation-aware and identity-tolerant. Signal forms drop the assumption: form(model) takes the model signal as its source of truth and does not keep its own copy, and for a Record<string, unknown> the field tree resolves to an index signature, so the rebuild step that causes those regressions stops existing. It does not replace the date-time picker's missing ControlValueAccessor — that component must keep working with [formControl], which Formly requires, so a proper accessor is its own ticket.
**Ticket:** *Evaluate signal forms for entry-components, piloting EntrySearchFilterComponent*

### 7. Still zone-based - Status: Separate ticket, after standalone.
**Evidence:** no `provideZonelessChangeDetection()` anywhere; the demo app is still
`platformBrowser().bootstrapModule(AppModule)` (`apps/demo-app/src/main.ts`) with `import 'zone.js'` in `polyfills.ts` and `zone.js ~0.16.0` in `package.json`. `NgZone.runOutsideAngular` survives in `entry-file-input.component.ts:41` and `markdown-viewer.component.ts:24` — under zoneless those calls are no-ops and the pattern is dead weight. `provideZonelessChangeDetection` is `@publicApi 20.2`, so it is stable. The signal migration on this branch is the prerequisite, so this is the natural next step rather than a parallel one.
**Ticket:** *Move the demo app to bootstrapApplication + zoneless, drop zone.js and NgZone usage*

### 8. Legacy Angular Material button API — includes a live defect - Status: The dead MatAnchor branch is gone in this PR — d0fe1cf. Confirmed at runtime, not just in typings: fesm2022/button.mjs has const MatAnchor = MatButton, so both injections resolved the same token. The M3 appearance / ThemePalette half is parked until a team asks.
**Evidence:** `entry-button.directive.ts` injects `MatButton` and `MatAnchor` as two separate
optional tokens and sets `.color` on whichever is present. In Material 22, `declare const MatAnchor: typeof MatButton` — **`MatAnchor` is an alias for `MatButton`**, the same token. Both injections resolve to one instance and the `matAnchor` branch is dead code.

Beyond that: the directive hand-maps MDC class strings (`['mdc-button', 'mdc-button--raised', 'mat-mdc-raised-button']`), which are internal Material implementation detail and will break on a Material minor, silently and visually. `MatButton` now exposes `appearance` via the `matButton` attribute (`'text' | 'filled' | 'elevated' | 'outlined' | 'tonal'`) plus a public `setAppearance()`. And `color` is not deprecated but its own doc says it "is supported in M2 themes only, it has no effect in M3 themes" — so `ThemePalette` in `entry-button-config.ts` and `spinner.component.ts` is a silent no-op for anyone on an M3 theme.
**Ticket:** *Migrate EntryButtonDirective to MatButton appearance and drop hand-written MDC classes*

### 9. `any` across `libs/` - Status: Separate ticket, last — 16 of the 30 occurrences sit in code the other tickets delete or retype.
**Evidence:** 30 occurrences. The raised comment covered SearchFilterBase<any>[] in entry-search-filter.component.ts, which is unknown on the branch now — but the same problem is in writeValue(value: any) / registerOnChange(fn: any) (entry-file-input.component.ts:175-183), TemplateRef<any> (entry-dialog.component.ts:43), SelectOption.value?: any, and effectively all of libs/entry-form/autocomplete/select-configuration.interface.ts.
**Ticket:** *Replace `any` with `unknown`/generics across entry-components and entry-form* — ties to #5, since several of these are what make an uncalled signal type-check.

### 10. `async` pipe and Observable-typed models inside signal components - Status: Separate ticket, in v22, after signal forms.
**Evidence:** `select-search-filter.component.html:9` still routes through `(searchFilter().options$ | async)` inside a `@let` ternary; `formly-autocomplete.component.html:1` does the same. The root cause is the model: `AutocompleteSearchFilter.search()` and `SelectSearchFilter.options$` are typed as `Observable`, which is what forces the `toObservable` + `switchMap` bridge in `autocomplete-search-filter.component.ts`. Converting the templates without converting the models just moves the seam. Partly subsumed by #6.
**Ticket:** *Move search-filter option sources from Observable to signals*

11. *ngIf in generated demo code - Status: Done in this PR — a637d49.

Evidence: product-edit-generated.component.html:8,11. This is Entry codegen output, so I checked the generator before editing anything: codegen already emits @if, which makes this a stale snapshot rather than a generator gap, and a regeneration the whole fix. The replacement matches what codegen's current default lane produces, and it also drops four hand-edited CSS classes that exist nowhere else in the repo.

That leaves the snapshot half fresh, which is why the rest is a ticket rather than more commits here: the .ts half is untouched since 2023 and still pre-signals — @Input, @Output, EventEmitter, OnInit, FormlyFieldConfig — and BB has no source assembly to regenerate from, so there is nothing here to re-run.
Ticket: Give the generated demo example a source of truth in the repo

The codegen-side item that came out of this one is separate, and not ours to schedule: Rendering/DisclaimerTemplateAppender.cs prepends @ts-nocheck, @ts-ignore and @ts-expect-error above every generated component (codegen PR 60). That is what blocks the ticket above — a regenerated snapshot would arrive with type checking switched off.

### 12. `FileLoadService` returns Observables that the consumer immediately unwraps - Status: Done in this PR — 6c0e978. httpResource is @publicApi 22.0 in @angular/common@22.1.3; the @experimental tag only appears in the stale Angular 21 in node_modules.
**Evidence:** `markdown-viewer.component.ts` wraps `loadDocumentationFile()` in `firstValueFrom` inside a `resource()` loader. `httpResource` (`@publicApi 22.0`) is the v22 shape for exactly this. Demo-app only, so low value — but it is the pattern the demo app is teaching consumers.
**Ticket:** low priority, or fold into #10.

### 13. Function-valued inputs on `EntryDialogComponent` - Status: Separate ticket, together with the unknown thread on the same file.
**Evidence:** `confirmAction` / `cancelAction` are `input<() => Observable<unknown>>`. Callbacks-as-inputs predate `output()` and signals, and this is the same design problem as the "convert to generic so we don't have `unknown` that can be overridden with whatever" thread on `entry-dialog.component.ts:51` — both are downstream of the dialog taking behaviour as an input rather than emitting an event. Already agreed as out of scope here; worth one ticket covering both rather than two.
**Ticket:** *Redesign EntryDialogComponent confirm/cancel as typed outputs*

libs/entry-components/common/directives/ng-control-accessor.directive.ts
jovanasta: Agreed on the destination, but it's a behavioural refactor spanning the picker, the no-op accessor and the ngModel sync, so it's going in its own ticket rather than riding along with the signals migration.

What did land in 8c368bd is the rationale and the removal path, recorded on `NgControlAccessorDirective` and on the picker's `hostDirectives`, so the reasoning isn't only in this conversation. Leaving this thread open to track the refactor itself.

libs/entry-components/common/directives/ng-control-accessor.directive.ts
jovanasta: Recorded in 8c368bd — the no-op-accessor rationale and the long-term removal path are now on `NgControlAccessorDirective` and on the picker's `hostDirectives`, rather than living only in this thread.

The refactor that actually deletes the three pieces is tracked on the sibling thread and goes to its own ticket.

libs/entry-components/permissions/permission.directive.ts
```diff
@@ -7,22 +7,61 @@ import { EntryPermissionService } from './permission.service';
...
+   * `only` wins when both aliases are bound, which the selector allows but no caller does.
+   *
+   * @remarks Denies by default. A binding that has not resolved yet - `[entryPermissionsOnly]="user()?.permissions"`
+   * while the profile loads - arrives as `undefined`, and for a permission gate the safe reading of
+   * "nothing to check against" is hidden, not visible.
+   *
+   * `hasPermissions` is not reactive, so this recomputes on input changes only - the same cadence
+   * the previous input setters ran at.
+   */
+  private readonly isHidden = computed(() => {
```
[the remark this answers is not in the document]
jovanasta: The documentation half is fixed in 8c368bd: `permissions/README.md` no longer shows the `*` form, and deny-by-default is now in §10 of the migration notes alongside `disableConfirm`. The directive also evaluates both aliases when both are bound now, instead of letting `only` win, which is documented in the same entry.

On `<div entryPermissionsOnly>` though — a valueless static attribute binds `''`, not `undefined`, so it never reaches the deny-by-default branch. On master that same attribute called the setter with `''` and `hasPermissions('')` threw, so I don't think behaviour changed for that case. Is there a variant you had in mind where the input genuinely arrives as `undefined` from a static attribute?

libs/entry-components/README.md
```diff
@@ -102,6 +103,176 @@ workspace-absolute imports, replace it with explicit `paths` entries — note th
...
++ });
+  ...
+  setServerSideValidationErrors(error, this.form);   // still compiles, now fully typed
+```
+
+`searchFilterForm` is a [`FormRecord`](https://angular.dev/api/forms/FormRecord) — a `FormGroup`
+whose keys are not known at compile time, which is exactly what a filter set is. `FormRecord`
+extends `FormGroup` at runtime, so `instanceof` checks and every method behave identically.
+
+`NgControlAccessorDirective.control` is the one narrowing. It hands back `AbstractControl` because a
```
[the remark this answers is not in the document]
jovanasta: I don't think those four are reachable by consumers: `entry-search-filter.module.ts` exports only `EntrySearchFilterComponent`, and the entry point's `public-api.ts` exports only that too. The four sub-components are declarations, so they can't be used in a consumer template or imported from the package, and a typed group can't be bound to them in the first place.

Have I missed a path where they're reachable? If so the rows go in.

libs/entry-components/common/directives/scroll-to-invalid-control.directive.ts
```diff
@@ -1,47 +1,46 @@
...
-
-    if (firstInvalidControl) {
-      firstInvalidControl.scrollIntoView({
-        behavior: 'smooth',
-        block: 'center' // vertical alignment
-      });
-    }
-  }
-}
+import { afterNextRender, DestroyRef, Directive, ElementRef, inject } from '@angular/core';
```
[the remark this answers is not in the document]
jovanasta: Confirmed — the blobs really did flip from CRLF to LF, and there's no `.gitattributes` while `.editorconfig` sets no `end_of_line`, so it will recur for every Windows contributor.

Doing it as its own PR the way you suggested, so the normalization commit doesn't bury this one. Leaving this open until that lands.

jovanasta:
Before merging, please double check that all breaking changes are documented.

libs/entry-components/common/directives/ng-control-accessor.directive.ts
typhoon41: Which thread? Which ticket? I though this one is achievable in this ticket.

libs/entry-components/common/directives/scroll-to-invalid-control.directive.ts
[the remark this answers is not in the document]
typhoon41: Let's not have separate PR for all the tiny things. There is gonna be a lot of them without it. When everything is reworked, just fix this so we can have it merged.

libs/entry-components/permissions/permission.directive.ts
```diff
@@ -1,28 +1,70 @@
...
+    // `style` on the same element, leaving `<div style="display: flex" entryPermissionsOnly>` visible.
+    effect(() => this.toggleVisibility(!this.isHidden()));
   }
 
   private readonly toggleVisibility = (show: boolean): void => {
     const element = this.elementRef.nativeElement;
 
-    if(!show) {
+    // `nodeType` rather than `instanceof HTMLElement`: those constructors are browser globals and
+    // undefined under a Node SSR renderer. Only the comment node produced by the unsupported
```
typhoon41: I don't think we care about so many comments in this class and in general.

typhoon41 (changes requested):

libs/entry-components/search-filter/entry-search-filter.component.ts
```diff
@@ -1,75 +1,112 @@
...
+    // input array rebuilds, so a filter set replaced once its options load is rendered and bound.
+    effect(onCleanup => {
+      const filters = this.searchFilters();
+      const currentValues = untracked(() => this.built().form.value);
+      const formatSubscriptions = new Subscription();
+      this.formatSubscriptions = formatSubscriptions;
+
+      // Values are carried over by key so a rebind does not discard what the user typed - callers
+      // do bind `[searchFilters]="getFilters()"`, which hands over a fresh array on every check.
+      this.built.set({ filters, form: this.toFormGroup(filters, currentValues) });
```
typhoon41: **[P1] Preserve control state when rebuilding search filters**

When a consumer binds a fresh array such as `[searchFilters]="getFilters()"`, every change-detection pass rebuilds the form with new controls. Only `.value` is copied, so dirty/touched/disabled state and validation errors are discarded; for example, errors applied with `setServerSideValidationErrors` can disappear on the next pass before `<entry-form-errors>` displays them. Reuse unchanged controls or preserve their full state.

libs/entry-components/common/directives/scroll-to-invalid-control.directive.ts
```diff
@@ -1,47 +1,56 @@
...
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
+import { ControlContainer } from '@angular/forms';
+import { fromEvent } from 'rxjs';
+import { NG_INVALID_CLASS } from '../constants';
+
+const FOCUSABLE_CONTROLS = 'input:not([type="hidden"]):not([disabled]),select:not([disabled]),textarea:not([disabled])';
```
typhoon41: **[P2] Focus invalid Material controls as well**

When the first invalid field is a `mat-select`, the matched element is a `mat-form-field` or `mat-select`, neither of which matches this native-only selector or contains a native `<select>`. `focusableWithin()` therefore returns `null`, leaving keyboard and screen-reader focus on the submit button despite the new focus behavior. Include Material/custom form-control focus targets or use a general focusability mechanism.

libs/entry-components/search-filter/entry-search-filter.component.ts
```diff
@@ -1,75 +1,112 @@
...
+  };
+
+  readonly toFormGroup = (
+    searchFilters: SearchFilterBase<unknown>[],
+    currentValues: Record<string, unknown> = {}
+  ): FormRecord => {
+    const group: Record<string, AbstractControl> = {};
+    searchFilters.forEach(searchFilter => {
+      const formControl = searchFilter.toFormControl();
+      if (searchFilter.key in currentValues) {
```
typhoon41: **[P2] Check only own keys when carrying values forward**

For a valid filter key inherited from `Object.prototype`, such as `constructor` or `toString`, this condition is true even on the initial empty `currentValues` object. The new control is then initialized with the inherited function instead of the filter's configured value. Use an own-property check such as `Object.hasOwn(currentValues, searchFilter.key)`.

typhoon41 (commented):
Codex re-review round (branch vs master) — three findings below as inline comments.

typhoon41 (approved):

libs/entry-components/common/directives/ng-control-accessor.directive.ts
StojanovicMilos: Reopening this to correct the mechanism, not the destination. All three pieces still come out together — the no-op accessor, this directive, and the `ngModel` sync. What changed is that the ticket should not implement `ControlValueAccessor` to do it.

`ControlValueAccessor` is the legacy custom-control contract in v22. The current one is `FormValueControl` from `@angular/forms/signals`: a `value = model<T>()` plus optional signal inputs (`errors`, `disabled`, `touched`, `dirty`, `invalid`, `required`, `min`, `max`) and a `touch` output.

The reason it got ruled out here was a mischaracterisation on my part, in the sibling thread: "signal forms would also remove it (`@angular/forms/signals` is public API in 22.0, with a `compat` bridge), but they are not the prerequisite". `signals/compat` is `compatForm`, which lets `FormControl` instances live inside a signal-forms model for a top-down migration. It has nothing to do with writing a custom control, and `FormValueControl` needs no bridge at all — reactive and template-driven forms drive it natively. So the sentence made the modern contract look like it cost us a signal-forms migration first, and it doesn't.

Verified on 22.1.0 rather than from the guide, with a throwaway control in the demo app. With no accessor present:

| what | result |
|---|---|
| `[formControl]` → child | `value` model receives the control's value |
| child → `[formControl]` | writing the model sets `control.value`, and `dirty` flips on both |
| `formControlName` + `Validators.required` | child receives `required=true`, `invalid=true`, `errors=[required]` |
| `touch` output | blur marks the bound control touched |
| `control.disable()` / `.enable()` | child's `disabled` input follows |
| `control.setValue()` | child's `value` model follows |
| **same component, but providing a no-op accessor** | **the model is never written** — the control stayed at its initial value |

Mechanically: detection is `hasModelInput(directiveDef, 'value')` in `@angular/core`'s `initializeCustomControlStatus` — a `model()` named `value`, nothing else, and it works on an NgModule-declared component. `NgControl.ngControlCreate` then takes the custom-control path whenever the element has no accessor, wiring the model to `control.setValue(v, {emitModelToViewChange: false})` + `markAsDirty()` + `viewToModelUpdate(v)`. That last call is exactly what `keepNgModelInSync` hand-rolls in this file, so the sync isn't replaced by anything — the framework already does it.

Three consequences for the ticket.

`errors` arriving as a bound input removes a problem the accessor route created and I had not spotted when we agreed it. An accessor that provides `NG_VALUE_ACCESSOR` cannot inject `NgControl` without a cyclic dependency, so `<mat-error entryDisplayControlValidation [control]="formControl">` would have had no control to bind and BP-1167 would have regressed. `EntryFileInputComponent`, which the ticket named as the pattern to copy, has no `mat-error` and so is silent on this.

Implementing both contracts is not an option, which is the last row of the table. `ngControlCreate` reads `if (hasCva || !host.customControl) return`, so providing an accessor makes the framework skip the custom-control path and the signal contract becomes dead code — silently, with no compile error.

There is no second breaking change later. One conversion serves reactive, template-driven and signal forms, so `formControlName` call sites never change and `[formField]` works afterwards with no further edit. That also means this is independent of the signal-forms work on the search filter rather than sequenced behind it.

Our peer range is already `^22.1.0`, so every consumer of 22.0.0 has this.

BP-1644 is retitled and rewritten accordingly, and BP-1642's breaking-set line with it.

libs/entry-components/common/directives/ng-control-accessor.directive.ts
StojanovicMilos: Answering your "in which exact step is time for it to go?" properly: the step is BP-1644, and the answer I gave on this thread was wrong about how.

The `compat` sentence above is the mistake — `signals/compat` is `compatForm`, a top-down migration aid, not how a custom control is written. `FormValueControl` needs no bridge and no signal-forms migration first; reactive and template-driven forms drive it natively, which I have now measured on 22.1.0. So the modern contract was available all along and is the one that actually answers "in the signal era".

Correction with the evidence is on the sibling thread where the decision was taken: https://github.com/enigmatry/entry-angular-building-blocks/pull/291#discussion_r3803884807

